PR #25, opened 2026-04-09
feat: num/caps lock on start

Kyza:
~~This PR adds Wayland's virtual keyboard protocol. It's off by default, and no client filter has been implemented. It can be toggled on/off live on config reloads.~~ It also adds support for turning numlock on by default.

```toml
[input.keyboard]
# num_lock = true            # num lock state on startup
# caps_lock = false          # caps lock state on startup
```

I'd like to also add scroll lock support, however, `ModifiersState` only has the other two.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -14,6 +14,12 @@
 # Each entry is passed to sh -c, so full shell syntax works (pipes, &&, env vars).
 # autostart = []
 
+# Disable numlock on startup.
+# numlock = false
+
+# Enable virtual keyboard protocol.
+# virtual_keyboard = true
+
 # Environment variables set before any clients launch.
 # Child processes (autostart, exec bindings) inherit these.
 # These override the compositor's built-in toolkit defaults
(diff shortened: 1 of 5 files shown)


malbiruk:
Thanks for the PR! 

The num_lock/caps_lock part looks good. However, I'd like to handle virtual keyboard separately. It needs proper client filtering rather than a config toggle, and the current Config::load() in the filter closure has performance issues.

Could you update this PR to only include the num_lock/caps_lock changes?

Also, the example config shows num_lock = false but the default is actually true. Example config shows default values unless the examples there are explicitly specified and marked with double comment ("# # ").

Kyza:
Thank you! If you could give me an example of how you'd like client filtering done, I could try to implement it myself. Otherwise I'm fine with just removing it from this PR, or making a new PR for it with those changes.

I didn't want to load the config in the filter, but removing/updating the protocol when it was disabled in the config was causing issues with crashes I didn't understand.

I'll update the example config when I get the time to work on this.

malbiruk:
Yeah, let's just remove virtual keyboard from this PR for now.

I checked out how niri handles it, they use the security context protocol (`wp_security_context_v1`) to mark sandboxed clients as restricted, then gate all privileged protocols (virtual keyboard, screencopy, etc.) behind that. I'll likely go the same route, so virtual keyboard would just be always-on with no config toggle.

PR #26, opened 2026-04-11
Implement Explicit Sync and NVIDIA stability quirks

wwmaxik:
This PR introduces support for Explicit Sync (linux-drm-syncobj-v1) and adds several stability 'quirks' specifically for NVIDIA GPUs to eliminate flickering and frame stuttering.

### Key Changes:
- **Smithay Upgrade**: Bumped smithay to the latest git main to enable the  protocol.
- **NVIDIA Detection**: Added automatic detection of NVIDIA GPUs via sysfs vendor ID (`0x10de`).
- **Legacy DRM API for NVIDIA**: Forced `SMITHAY_USE_LEGACY=1` for NVIDIA devices. The Atomic Modesetting API in Smithay can sometimes cause 'one-frame-behind' flickers on NVIDIA; Legacy DRM provides a much smoother experience.
- **Manual Fence Synchronization**: Implemented an explicit wait for GPU fences (`SyncPoint::wait()`) in the rendering loop for NVIDIA devices to ensure buffers are fully ready before calling page flip.
- **Scanout Lockdown**: Disabled direct scanout (`FrameFlags::empty()`) for NVIDIA to prevent 'flicker to black' artifacts, ensuring all frames go through EGL composition.
- **API Migration**: Updated the codebase to match latest Smithay API changes (unified `DndGrabHandler`, `ToplevelCachedState`, and updated `RenderElement` traits).

These changes significantly improve the user experience on NVIDIA GTX 10-series and newer cards (tested on GTX 1650).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,23 +32,21 @@ dependencies = [
 
 [[package]]
 name = "android-activity"
-version = "0.6.0"
+version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ef6978589202a00cd7e118380c448a08b6ed394c3a8df3a430d0898e3a42d046"
+checksum = "0f2a1bb052857d5dd49572219344a7332b31b76405648eabac5bc68978251bcd"
 dependencies = [
  "android-properties",
  "bitflags 2.11.0",
  "cc",
- "cesu8",
  "jni",
- "jni-sys",
  "libc",
  "log",
  "ndk",
  "ndk-context",
  "ndk-sys",
  "num_enum",
- "thiserror 1.0.69",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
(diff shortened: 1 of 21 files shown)


malbiruk:
Thanks for the PR! I appreciate the effort, and there are things I'd definitely want to take.                                                  

This PR is pretty big though and touches multiple concerns which makes it hard to merge as-is.
                                                
**Smithay API migration** -- that's the right move, I'd like to take it.                                                                  

**NVIDIA quirks** -- I think that auto-detection approach may be fragile (especially in multi-GPU setups) and maybe not all users would like to use all optimization quirks at once. I'd expose debug config flags (`force_legacy_drm`, `wait_for_frame_completion`, `disable_direct_scanout`) and let users opt in.                                                                       

**Blur half-res optimization** -- I'd take this, could go with the smithay migration PR or standalone.

**Spring animations + entry animations** -- interesting feature but out of scope here. Happy to look at it as a separate PR if you want to contribute it.

wwmaxik:
Closed in favor of split PRs #27, #28, and #29 based on feedback.

PR #27, opened 2026-04-12
Smithay API migration and Blur performance optimization

wwmaxik:
This PR updates Smithay to the latest git main and introduces a major optimization for the Kawase blur effect.

### Key Changes:
- **Smithay Upgrade**: Migration to the latest API (DndGrabHandler, ToplevelCachedState).
- **Blur Optimization**: Implemented 2x downsampling for the Kawase blur and background FBO. This reduces the pixel processing load by 4x, significantly improving performance on high-resolution displays.

PR #28, opened 2026-04-12
Add backend configuration flags for hardware stability quirks

wwmaxik:
This PR adds manual configuration flags to the `[backend]` section to address stability issues on certain hardware (specifically NVIDIA).

### Added Flags:
- `force_legacy_drm`: Forces the Legacy DRM API (instead of Atomic).
- `wait_for_frame_completion`: Explicitly wait for GPU fences before flipping.
- `disable_direct_scanout`: Disables direct scanout to ensure EGL composition.

It also includes smart defaults that automatically enable these quirks if an NVIDIA GPU is detected, while allowing users to override them as requested in the previous PR review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,23 +32,21 @@ dependencies = [
 
 [[package]]
 name = "android-activity"
-version = "0.6.0"
+version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ef6978589202a00cd7e118380c448a08b6ed394c3a8df3a430d0898e3a42d046"
+checksum = "0f2a1bb052857d5dd49572219344a7332b31b76405648eabac5bc68978251bcd"
 dependencies = [
  "android-properties",
  "bitflags 2.11.0",
  "cc",
- "cesu8",
  "jni",
- "jni-sys",
  "libc",
  "log",
  "ndk",
  "ndk-context",
  "ndk-sys",
  "num_enum",
- "thiserror 1.0.69",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -211,9 +209,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.56"
+version = "1.2.60"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "aebf35691d1bfb0ac386a69bac2fde4dd276fb618cf8bf4f5318fe285e821bb2"
+checksum = "43c5703da9466b66a946814e1adf53ea2c90f10063b86290cc9eb67ce3478a20"
 dependencies = [
  "find-msvc-tools",
  "jobserver",
@@ -222,16 +220,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "cesu8"
-version = "1.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6d43a04d8753f35258c91f8ec639f792891f748a1edbd759cf1dcea3382ad83c"
-
-[[package]]
 name = "cfg-expr"
-version = "0.20.6"
+version = "0.20.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "78cef5b5a1a6827c7322ae2a636368a573006b27cfa76c7ebd53e834daeaab6a"
+checksum = "3c6b04e07d8080154ed4ac03546d9a2b303cc2fe1901ba0b35b301516e289368"
 dependencies = [
  "smallvec",
  "target-lexicon",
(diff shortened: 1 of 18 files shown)


malbiruk:
The config flags are exactly what I wanted, thanks!

One thing though -- I'd prefer defaults be `false` across the board rather than auto-detected from sysfs. Users opt in via config. This way there's no runtime sysfs detection, and users can mix/match quirks for their specific setup. I'll document which flags to enable for NVIDIA.

With opt-in defaults, the `is_nvidia` field on `DriftWm` isn't needed anymore -- code can check `config.backend.wait_for_frame_completion` etc. directly.

wwmaxik:
Sounds like a plan. I'll switch all defaults to false and remove the sysfs detection logic to keep it clean. I'll also drop the is_nvidia field and update the checks to reference the config flags directly. I'll get to it this evening and push the commit

malbiruk:
Thanks for splitting this! But a few things from the last round still need fixing before I can merge:

1. Drop is_nvidia_gpu() sysfs detection.
The compositor shouldn't be sniffing GPU vendor from /sys/class/drm. Users who need the quirks can just set the flags in their config.

2. Default all three flags to false.
force_legacy_drm: raw.force_legacy_drm.unwrap_or(nvidia), ...
Should be unwrap_or(false). Fully opt-in. Add a comment in config.example.toml like "enable if you're on NVIDIA and see flickering or crashes."

3. Remove the is_nvidia field on DriftWm.
With detection gone there's nothing to put in it. Right now it's assigned from force_legacy_drm, which makes the name misleading.

4. Move the env var stuff out of init_udev.
unsafe { std::env::set_var(...) } is problematic there: in Rust 2024 it's UB when other threads are running, and by the time udev init runs the event loop is already up. On top of that, __GL_GSYNC_ALLOWED / __GL_VRR_ALLOWED / __GL_MaxFramesAllowed are read by NVIDIA's driver at init. Setting them after the process started probably does nothing. These belong in the driftwm-session wrapper script, or just documented so users export them before launching.

5. Drop AnimationConfig from this PR.
The struct and parser are here but the code that uses them is in #29. Better to land it together in #29.

6. Maybe split force_legacy_drm?
Right now it toggles five things at once (legacy DRM, G-Sync, VRR, frame cap, NVD backend). If a user says "this fixed my flicker," we won't know which one actually helped. Not a blocker if you'd rather keep it simple.

wwmaxik:
All feedback addressed:

✅ **Removed GPU vendor detection** - Dropped `is_nvidia_gpu()` sysfs sniffing entirely
✅ **All flags default to false** - Fully opt-in, no auto-enabling
✅ **Removed `is_nvidia` field** - No longer stored in DriftWm state
✅ **Removed env var manipulation** - Dropped all `unsafe { std::env::set_var(...) }` calls from init_udev. Users should set `__GL_GSYNC_ALLOWED`, `__GL_VRR_ALLOWED`, `__GL_MaxFramesAllowed`, and `NVD_BACKEND` in their session wrapper or shell profile before launching driftwm
✅ **Removed AnimationConfig** - Struct and parser removed (will land together with PR #29)
✅ **Added config documentation** - Detailed comments in config.example.toml explaining when to enable each flag

The backend flags are now purely opt-in configuration with no runtime detection or environment manipulation.

malbiruk:
Since you're stepping back from the blur optimization #27 , can you revert those render.rs changes from this PR? Then I can merge the smithay migration + backend flags.

malbiruk:
Also one more thing -- `force_legacy_drm` is now a dead config option after the env var block was removed. It's parsed and stored but nothing reads it.

Setting it to true does nothing. Could you remove it entirely? Users who need legacy DRM can set SMITHAY_USE_LEGACY=1 in their session wrapper script alongside the __GL_* vars -- same pattern as the other env vars you documented in config.example.toml.

wwmaxik:
Updated PR to address feedback:

✅ **Removed force_legacy_drm** - This was dead code after the env var block was removed. Users who need legacy DRM should set `SMITHAY_USE_LEGACY=1` in their session wrapper script before launching driftwm.

✅ **Updated config.example.toml** - Added documentation for all NVIDIA-related environment variables that should be set in the session wrapper:
- `SMITHAY_USE_LEGACY=1` (legacy DRM API)
- `__GL_GSYNC_ALLOWED=0`
- `__GL_VRR_ALLOWED=0`
- `__GL_MaxFramesAllowed=1`
- `NVD_BACKEND=direct`

✅ **Kept two working flags**:
- `wait_for_frame_completion` - Wait for GPU fences before page flip
- `disable_direct_scanout` - Force EGL composition

Ready to merge!

wwmaxik:
By the way, I noticed you're from Russia - would you prefer to communicate in Russian? :)

If you'd like to discuss details or communicate faster, I can share my Telegram contact. Or we can continue here in PRs, whatever works best for you!

malbiruk:
Could you also revert blur optimization (restore src/render.rs to match main)? That's the last blocker before I can merge this.

I'd prefer to keep things in English here on the PRs, that way other contributors or users browsing the repo later can follow the discussions. Public PR history is useful as a record.

wwmaxik:
Sure, I'll revert the blur optimization and restore src/render.rs to match main. And no problem - we'll keep discussions in English here for the PR history. I'll take care of this tomorrow around 5 PM GMT+7.

malbiruk:
Also rebase on current main when you'll be at it please

wwmaxik:
✅ **Reverted render.rs to match main** - Removed all blur optimization changes as requested

✅ **Rebased on current main** - Branch is now up to date with latest upstream/main

Ready to merge! The PR now only contains:
- Backend configuration flags (wait_for_frame_completion, disable_direct_scanout)
- Documentation for NVIDIA env vars in config.example.toml
- No render.rs changes
- No force_legacy_drm (users should set SMITHAY_USE_LEGACY=1 in session wrapper)

PR #33, opened 2026-04-14
Add time uniform support for animated background shaders

### LINKED ISSUE #2: Animated shader backgrounds
At present, the backgrounds are static and are only re-rendered when the viewport moves. This is fine and makes a lot of sense, but it'd be fun to be able to animate the backgrounds with time (e.g. have pulsating waves, rain droplets, directional indicators, etc). Imagine all of the fun demos one could port from shadertoy!

It may also be worth feeding in additional information to the shader to enable other forms of animation (e.g. what shadertoy does), but this isn't super essential. It could enable interesting things like changing the background with the time or weather, but you'd need some kind of mechanism to feed that information in.

wwmaxik:
This PR adds support for animated background shaders by providing a `u_time` uniform.

## Changes
- Added `u_time` uniform to background shaders (time in seconds since compositor start)
- Update time uniform on every frame for smooth animations
- Added example animated wave shader (`extras/wallpapers/animated_waves.glsl`)
- Documented shader uniforms in `config.example.toml`

## Available Shader Uniforms
- `u_camera`: vec2 - camera position in canvas coordinates
- `u_time`: float - time in seconds since compositor start
- `u_size`: vec2 - viewport size in pixels

## Use Case
Enables animated backgrounds without requiring external tools or video playback. Users can create dynamic effects like waves, particles, color cycling, or any time-based animation directly in GLSL.

## Example
See `extras/wallpapers/animated_waves.glsl` for a simple animated wave effect.

Clean implementation with minimal code changes - just adds the time uniform and updates it each frame.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -123,6 +123,13 @@
 # Examples:
 # # shader_path = "~/.config/driftwm/bg.glsl"   # custom GLSL fragment shader
 # # tile_path = "~/.config/driftwm/tile.png"     # tiled image (mutually exclusive with shader)
+#
+# Custom shaders receive these uniforms:
+# - u_camera: vec2 - camera position in canvas coordinates
+# - u_time: float - time in seconds since compositor start (for animations)
+# - u_size: vec2 - viewport size in pixels
+#
+# Example animated shader: extras/wallpapers/animated_waves.glsl
 
 # Keyboard bindings: "Modifier+...+Keysym" = "action [arg]"
 # Merges with defaults. Use "none" to unbind a default binding.
--- a/extras/wallpapers/animated_waves.glsl
+++ b/extras/wallpapers/animated_waves.glsl
@@ -1,0 +1,25 @@
+// Animated wave shader example
+// Available uniforms:
+// - u_camera: vec2 - camera position in canvas coordinates
+// - u_time: float - time in seconds since compositor start
+
+precision mediump float;
+
+varying vec2 v_coords;
+uniform vec2 size;
+uniform float alpha;
+uniform vec2 u_camera;
+uniform float u_time;
+
+void main() {
+    vec2 screen_pixel = v_coords * size;
+    vec2 canvas_pos = screen_pixel + u_camera;
+    vec2 pos = canvas_pos / 100.0;
+    
+    // Animated waves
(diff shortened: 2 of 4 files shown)


malbiruk:
Good PR overall, backward compatibility checks out too -- static background shaders work without changes.

One issue though: animated shaders won't actually animate in udev mode (TTY) if no client / cursor movements create damage. Frames only render when state changes, so u_time advances but the frame stays stale until the user interacts. Works in nested winit because the 60fps timer always fires, but breaks in production.

Can you schedule continuous redraws when an animated shader is detected?
Simplest: check if shader_source.contains("u_time") at shader load, and if so mark the output dirty after each render. That way VBlank drives the animation at the display's refresh rate.

wwmaxik:
Got it, thanks for the diagnosis. I haven't reproduced it myself, but I see the issue. I'll implement the fix - checking for u_time on shader load and scheduling continuous redraws via VBlank when needed. Will get to it tomorrow around 5 PM GMT+7.

malbiruk:
To easily reproduce, you can move viewport to empty space and not move cursor -- there should be no blinking cursors etc. -- nothing to trigger redraws. Also driftwm should be started from TTY (not nested session). In that case the background becomes static.

wwmaxik:
Thanks for the steps. I'll reproduce that way first, then implement the fix.

malbiruk:
Sorry, can you rebase on main and resolve conflicts please?

Your PR #28  inspired me to investigate what was the cause of those black 1px flickers (I noticed them at some zoom levels to, and I'm on AMD), so I added a proper fix and then refactored render module into submodules

wwmaxik:
Sure, I'll get the rebase done within half an hour. Also, that's great news about the flicker fix! I'll take a look at the new submodules

wwmaxik:
Rebased on main and resolved conflicts. Changes applied:
- Updated BG_UNIFORMS to include u_time uniform
- Modified update_background_element() to pass time uniform
- Modified init_background() to detect animated shaders and initialize with time uniform
- Added background_is_animated flag to RenderCache
- Modified post_render() to schedule continuous redraws for animated shaders

All changes have been adapted to the new render module structure (render/ subdirectory). Build passes successfully.

PR #44, opened 2026-04-19
Optimize blur: per-window background caching

### LINKED ISSUE #42: Blur causes 90% GPU usage (NVIDIA)
Context: Gentoo Linux, driftwm built from source code, alacritty terminal, mostly default config, nvidia RTX2060s + 590.48.01 drivers, terminal blur from example config enabled. 

Config snippet:
```
#Blurred terminal
[[window_rules]]
app_id = "Alacritty"
opacity = 0.85
blur = true
```
The issue itself: when default blur for terminal is enabled and the terminal window is visible on the screen GPU usage jumps up to 95%. Depends on how much terminal windows you opened. Normal Idle usage for me: 10-30%

10-30% idle is also kinda abnormal, but somewhat ok for experimental software

No blur GPU usage:

Image

GPU usage with blur:
Image

wwmaxik:
## Summary

Fixes #42 by implementing per-window background fingerprinting for blur cache invalidation.

## Problem

Previously, blur cache was invalidated on **any** surface commit (via global `blur_scene_generation`), causing blur to recompute even when the background behind a specific window hadn't changed. This led to:
- High GPU usage scaling linearly with number of blurred windows
- 90%+ GPU usage reported with multiple blurred terminals on NVIDIA

## Solution

Replace global scene generation tracking with per-window background hash:
- Each `BlurCache` now stores `last_background_hash` 
- Hash computed from elements behind each window + window rect
- Blur only recomputes when that specific window's background changes

## Changes

- Add `last_background_hash` field to `BlurCache`
- Implement `hash_background_elements()` to fingerprint background
- Move `behind_starts` calculation before cache invalidation check
- Replace `scene_generation` check with background hash comparison

## Test Results (NVIDIA GTX 1650)

**Before:**
- 1 blurred window: ~30-37% GPU
- Multiple blurred windows: scales linearly (90%+ reported)

**After:**
- 1 blurred window: ~10-38% GPU (varies with animated background)
- 6 blurred windows: ~29-37% GPU (stable, no scaling!)

**Key improvement:** GPU usage no longer scales with number of blurred windows.

## Notes

- Remaining ~30% GPU usage is from animated background shader (60 FPS rendering)
- With static shader: GPU drops to 2-3% idle
- Debug logs added to track invalidation reasons (can be removed if needed)

### DIFF
--- a/src/render/blur.rs
+++ b/src/render/blur.rs
@@ -1,16 +1,42 @@
 use smithay::backend::allocator::Fourcc;
 use smithay::backend::renderer::element::texture::TextureRenderElement;
-use smithay::backend::renderer::element::Kind;
+use smithay::backend::renderer::element::{Element, Kind};
 use smithay::backend::renderer::gles::{
     GlesError, GlesRenderer, GlesTexProgram, GlesTexture, Uniform, UniformName, UniformType,
 };
 use smithay::output::Output;
 use smithay::utils::{Physical, Rectangle, Size, Transform};
+use std::collections::hash_map::DefaultHasher;
+use std::hash::{Hash, Hasher};
 
 use super::OutputRenderElements;
 
 static BLUR_DOWN_SRC: &str = include_str!("../shaders/blur_down.glsl");
 static BLUR_UP_SRC: &str = include_str!("../shaders/blur_up.glsl");
+
+/// Compute a hash of render elements behind a window for blur cache invalidation.
+/// This allows per-window background tracking instead of global scene generation.
+fn hash_background_elements(elements: &[OutputRenderElements], window_rect: Rectangle<i32, Physical>) -> u64 {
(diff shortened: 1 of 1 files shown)


malbiruk:
Thanks for working on this! I tested it on my laptop and on idle the per-window background caching optimization doesn't show a measurable improvement: 1 blurred terminal sits at ~2% GPU usage, and 6 blurred terminals at ~7% -- both with and without the change. I suspect the original 30% and 90% GPU from #42 might be NVIDIA-specific rather than a cache invalidation issue or something else I'm missing...

Could you try reproducing the original issue on an unmodified main branch on your hardware?

malbiruk:
Active usage seems to improve a bit though, so I'll merge with follow-up commit consolidating the original needs_recompute block with the new one.

PR #50, opened 2026-04-20
backend/udev: make NVIDIA legacy DRM rendering conservative

komarukomaru:
## Summary

Fix rendering artifacts on NVIDIA when driftwm runs on the legacy DRM path.

On an RTX 3060 dual-monitor setup, driftwm showed browser/video flicker, frame stutter, and visible rendering corruption. The same
machine under GNOME Wayland did not reproduce the issue, which pointed to driftwm's compositor path rather than a general Wayland/
NVIDIA issue.

## What changed

- detect NVIDIA GPUs explicitly in the udev backend
- keep full DMA-BUF feedback for Wayland clients
- restrict Xwayland surfaces to implicit-modifier DMA-BUF feedback only
- disable `linux-drm-syncobj-v1` advertisement on legacy DRM
- force full redraws on NVIDIA + legacy DRM by resetting buffer ages every frame
- avoid queueing empty frames

## Rationale

This path was too optimistic for NVIDIA on legacy DRM:
- explicit sync advertisement on the legacy path was not conservative enough
- buffer-age reuse / partial damage was likely contributing to flicker and corruption
- Xwayland needed stricter DMA-BUF feedback than native Wayland clients

Mutter has accumulated several NVIDIA-specific fixes in the same general area: synchronization, sharable surfaces, and damage handling.
This change follows the same direction by making the problematic path more conservative.

## Tradeoffs

This intentionally favors correctness over efficiency on the affected path.

Possible downsides:
- higher GPU usage
- more full-frame redraw work on NVIDIA + legacy DRM

Atomic DRM behavior is left unchanged.

## Testing

Tested on:
- NVIDIA GeForce RTX 3060
- dual-monitor setup
- `DP-1` at `1920x1080@165`
- `HDMI-A-1` at `1280x1024@60`

Validation:
- `cargo check`
- `cargo build --release`
- `~/.local/bin/driftwm --check-config`

Observed result:
- black-window regression fixed
- rendering corruption fixed
- flicker significantly improved on the legacy NVIDIA path

Если хочешь покороче, можно такой вариант:

## Summary

Make the NVIDIA legacy DRM path more conservative to avoid rendering corruption and flicker.

## Changes

- restrict Xwayland DMA-BUF feedback to implicit modifiers
- disable `linux-drm-syncobj-v1` on legacy DRM
- force full redraws on NVIDIA + legacy DRM
- skip queueing empty frames

## Why

This fixes rendering artifacts seen on an RTX 3060 dual-monitor setup under driftwm, while the same machine behaved correctly under
GNOME Wayland.

## Validation

- `cargo check`
- `cargo build --release`
- `~/.local/bin/driftwm --check-config`

### DIFF
--- a/src/backend/udev.rs
+++ b/src/backend/udev.rs
@@ -1,6 +1,6 @@
 use std::cell::RefCell;
 use std::collections::HashMap;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::rc::Rc;
 
 use smithay::{
@@ -68,6 +68,26 @@ struct SurfaceData {
     make: String,
     model: String,
     serial_number: String,
+}
+
+fn is_nvidia_gpu_path(path: &Path) -> bool {
+    let Some(card_name) = path.file_name().and_then(|n| n.to_str()) else {
+        return false;
+    };
+    let vendor_path = PathBuf::from("/sys/class/drm")
+        .join(card_name)
+        .join("device/vendor");
+    std::fs::read_to_string(vendor_path)
+        .map(|s| s.trim().eq_ignore_ascii_case("0x10de"))
+        .unwrap_or(false)
+}
+
+fn implicit_modifier_formats(formats: &[Format]) -> Vec<Format> {
+    formats
+        .iter()
+        .copied()
+        .filter(|f| f.modifier == Modifier::Invalid)
+        .collect()
 }
 
 /// Opaque handle to udev backend device data. Returned by init_udev,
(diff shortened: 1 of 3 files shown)


malbiruk:
Thanks for the detailed writeup!

Before we go further with this PR, could you try the existing `[backend]` opt-in quirks in config.example.toml?

```
[backend]
disable_direct_scanout = true
wait_for_frame_completion = true
```

These target exactly this class of NVIDIA issue:
- disable_direct_scanout composites every buffer through EGL instead of scanning out directly -- a superset of the XWayland implicit-modifier workaround in this PR.
- wait_for_frame_completion handles NVIDIA's early page-flip completion, likely the flicker/stutter cause.

If those fix it on your setup, we'd prefer keeping these as explicit opt-ins rather than auto-detecting NVIDIA and branching render paths -- driftwm has been deliberately avoiding vendor-gated code.

Force-full-redraws is also a big permanent GPU cost to paper over buffer-age issues rather than understand them.

One piece of the PR I think is worth pursuing independently: not advertising `linux-drm-syncobj-v1` when we've fallen back to legacy DRM. That's not NVIDIA-specific and seems wrong for any GPU. Happy to review that as a small standalone PR.

komarukomaru:
Thanks, I had already tested with both of these enabled:

[backend]
disable_direct_scanout = true
wait_for_frame_completion = true

On my setup they did not resolve the issue completely: I still saw flicker / stutter / rendering corruption
on the legacy DRM path.

That said, I understand the preference to avoid vendor-specific branching in driftwm. I agree this PR is
too broad in its current form.

I'll close this one and send a smaller follow-up PR just for disabling linux-drm-syncobj-v1 advertisement
on legacy DRM, since that part seems correct independent of GPU vendor.

PR #59, opened 2026-04-25
feat: per-window capture via hyprland-toplevel-export-v1

lama2923:
Implements hyprland-toplevel-export-v1 (v2) so xdg-desktop-portal-hyprland can share individual windows. Tested: OBS, Discord window capture.


malbiruk:
Hey man, thanks for the PR!

Unfortunately it doesn't work on my system. Reproducibility details and Claude's analysis below.

Tested on Fedora 43, `xdg-desktop-portal-hyprland 1.3.11`. **Per-window capture is broken end-to-end** — Google Meet in Brave shows the picker but never streams; OBS additionally crashes in `g_variant_builder_add_value()` (confirmed via `coredumpctl`) — independent OBS bug, but our code triggers it.

xdph journal:

```
[LOG]  | Got interface: hyprland_toplevel_export_manager_v1 (ver 2)
[LOG] [screencopy] Registered for toplevel export
[LOG] [sc] Selection: /window:1867580848
[LOG] [screencopy] Start:
[ERR] [screencopy] Couldn't obtain a format from dma
[LOG] [screencopy] Stream destroyed
```

The error fires at [`Screencopy.cpp:300`](https://github.com/hyprwm/xdg-desktop-portal-hyprland/blob/main/src/portals/Screencopy.cpp#L300) when `frameInfoDMA.fmt == DRM_FORMAT_INVALID` — xdph's `setLinuxDmabuf` callback for `CCHyprlandToplevelExportFrameV1` ([`Screencopy.cpp:543`](https://github.com/hyprwm/xdg-desktop-portal-hyprland/blob/main/src/portals/Screencopy.cpp#L543)) never fires with a valid format. SHM fallback isn't an option — xdph has a `// todo: blocks shm` comment confirming this protocol requires dmabuf.

Ruled out: `Fourcc::Xrgb8888 as u32` does evaluate to `0x34325258` (`DrmFourcc` is `#[repr(u32)]`), and the global binds fine.

Most likely causes:

1. **Handle ID mismatch in `surface_for_handle_id`** (`foreign_toplevel.rs:466`) — compares `&h.id() == handle_id` against server-side `ZwlrForeignToplevelHandleV1` instances. The `handle_id` from `CaptureToplevelWithWlrToplevelHandle` is the wire ObjectId xdph sent; these need to match per-client and may not.
2. **Dispatch isn't reached** — `client_is_unrestricted` filter on `can_view` may reject xdph, or a `delegate_*` macro routing issue.

@lama2923 — since this works for you on OBS/Discord, it'd help to know your setup so we can isolate the divergence:

1. Distro + xdph version
2. GPU (`lspci | grep VGA`)
3. Anything in `~/.config/xdg-desktop-portal/portals.conf`, or relying on the PR's system file?

If you can also add `tracing::info!("toplevel export: handle={handle_id:?} window_found={}", window.is_some());` after line 217 in `hyprland_toplevel_export.rs` and re-test with `RUST_LOG=info`, comparing what you log vs. what I log will tell us which of the two causes above is actually it (or whether the dispatch differs across xdph versions).

Happy to run further diagnostics once we know what to look for.

lama2923:
Understood the problem. I've corrected the implementation, fixed some dependencies, and properly added xdg-desktop-portal-hyprland.

Can you give it another go and check the build/execution?

@malbiruk

malbiruk:
Thank you, now the per-window screensharing works, but xdph's UX bugs me more than I expected (rough picker, double-pops in Chromium Google Meet, OBS crashes on first attempt) and none of it's fixable from our side. 

Meanwhile niri's docs recommend [`xdg-desktop-portal-gnome`](https://github.com/YaLTeR/niri/blob/main/docs/wiki/Screencasting.md) instead — it's compositor-agnostic and gives much nicer UX. We're two protocols away from supporting it: `ext-foreign-toplevel-list-v1` and `ext-foreign-toplevel-image-capture-source-manager-v1`. The smithay reexports exist and most of it mirrors our existing `foreign_toplevel.rs`.

So I'm leaning toward going that route instead of merging this. The code is solid but xdph locks us into UX problems that won't get better, and the alternative's genuinely nicer.

Sorry that I thought about it after the work was done. You can implement `ext-*` protocols if you want (your bind-time enumeration changes carry over), or I get to it eventually when I have time. No pressure either way.

lama2923:
I can't contribute to this PR right now, sorry. No point in leaving it open just to let it become trash. If someone wants to take it over, they can, but I'm closing it for now.

PR #64, opened 2026-04-29
Xwayland satellite

malbiruk:


malbiruk:
Merged as 6fafb06 (squash-merged locally).

PR #73, opened 2026-05-06
fix: render drag-and-drop cursor icon + add Unix socket IPC

### LINKED ISSUE #70: IPC
### Discussed in https://github.com/malbiruk/driftwm/discussions/67

<div type='discussions-op-text'>

<sup>Originally posted by **shananiki** May  2, 2026</sup>
Is there any chance it will support plugins and commands via IPC?</div>

### LINKED ISSUE #71: Drag-and-drop cursor icon not visible
## Description
When dragging files from Dolphin (or any app) to another window, the drag cursor icon is invisible. The cursor stays as a default pointer — you can't see what you're dragging.

## Root cause
In `handlers/mod.rs`, the `dnd_requested` handler receives the icon surface but discards it with `_icon`:
```rust
fn dnd_requested(&mut self, source: S, _icon: Option\<WlSurface\>, ...)
```
The icon is never rendered, making drag-and-drop operations blind.

## Expected behavior
The drag icon (file thumbnail, text preview, etc.) should appear at the cursor position during drag-and-drop.

### LINKED ISSUE #72: IPC: Unix socket for external control (camera, zoom, layout, state)
## Description
There's no way to control driftwm from external tools (shell, scripts, widgets). The state file at `$XDG_RUNTIME_DIR/driftwm/state` is read-only — tools can only poll it, not send commands.

## Proposed solution
A Unix socket at `$XDG_RUNTIME_DIR/driftwm/ipc.sock` that accepts text commands:

| Command | Action |
|---------|--------|
| `camera x y` | Pan viewport to canvas coordinates |
| `camera` | Get current viewport center |
| `zoom 0.5` | Set zoom level |
| `zoom` | Get current zoom |
| `layout us,ru` | Switch keyboard layout |
| `layout` | Get active layout |
| `state` | Dump full state file |
| `quit` | Graceful shutdown |

## Why
- **Desktop shells** (waybar, noctalia) need real-time state without polling
- **Keyboard layout switching** — currently impossible without this
- **Canvas navigation scripts** — pan/zoom from external tools
- **Health monitoring** — CPU/GPU stats, frame timing

Related: #70

youssefvdel:
Closes #70, closes #71, closes #72

### 1. Drag-and-drop cursor (fixes #71)

The DnD icon surface was received but discarded (`_icon`). Now:
- Stored on `DriftWm.dnd_icon` during the grab
- Rendered at pointer position in `build_cursor_elements()`
- Cleared when the grab ends via `DndGrabHandler::dropped`

### 2. Unix socket IPC (fixes #70, #72)

New module `src/ipc.rs` — creates `$XDG_RUNTIME_DIR/driftwm/ipc.sock`:

```bash
# Camera
echo "camera 500 300" | nc -U $XDG_RUNTIME_DIR/driftwm/ipc.sock

# Zoom  
echo "zoom 0.5" | nc -U $XDG_RUNTIME_DIR/driftwm/ipc.sock

# State
echo "state" | nc -U $XDG_RUNTIME_DIR/driftwm/ipc.sock

# Quit
echo "quit" | nc -U $XDG_RUNTIME_DIR/driftwm/ipc.sock
```

Integrated into the calloop event loop via `Generic\<OwnedFd\>`. One command per connection, responds then closes.

## Files changed
| File | Change |
|------|--------|
| `src/ipc.rs` | New — IPC module (162 lines) |
| `src/main.rs` | +8 — register IPC source in event loop |
| `src/handlers/mod.rs` | +4 — store/render/clear DnD icon |
| `src/render/mod.rs` | +18 — render DnD icon at pointer |
| `src/state/mod.rs` | +1 — dnd_icon field |
| `src/state/init.rs` | +1 — initialize dnd_icon |


Kyza:
It appears that your `graphify-out` directory was included in your commits, adding ~38,000 lines of changes. Is this intentional?

youssefvdel:
Done — graphify-out was included by accident in the initial push. I've added it to .gitignore and force-pushed both branches clean. The diff should now show only the actual code changes (~200 lines for the DnD/IPC PR). @Kyza

malbiruk:
Thanks for tackling all three of these! The DnD fix is great -- clean and minimal. The IPC and placement parts have some issues I'd like to work through though, so I'd prefer to split this into separate PRs.

**Could you split out the DnD fix into its own PR?** It's a clear win and I can merge it right away.

**On IPC — a few things:**

1. **Blocking `read_line` is a DoS vector.** The stream is set to blocking and read synchronously on the main event loop thread. Any local process can freeze the compositor:
   ```bash
   exec 3<>/run/user/$UID/driftwm/ipc.sock; sleep infinity
   ```
   The fix is to keep the listener nonblocking, wrap each accepted stream into calloop's async I/O (`event_loop.adapt_io(stream)`), schedule a per-client async task, and `read_until(b'\n').await`. That way a stuck client never stalls the compositor.

2. **`layout` command doesn't actually switch layouts.** Setting `data.active_layout = args` only stomps a cached status field that gets overwritten from XKB on the next keypress (`src/input/keyboard.rs:146`). To really switch layouts you'd need to drive the seat keyboard's XKB state — and we don't have a "switch layout" action wired up at all yet. Let's drop the write side of `layout` from this PR; once a real switch-layout action exists, IPC can route through it. Read side (`layout` with no args) is fine to keep.

3. **Listener fd handling is hacky.** Capturing `RawFd` and reconstructing `UnixListener::from_raw_fd` every callback (leaked via `ManuallyDrop`) works but is fragile. The clean pattern: `UnixListener::bind` once, wrap in `Generic::new(listener, ...)`, and the calloop callback receives `&UnixListener` directly — call `socket.accept()` on it.

**On placement — let's drop the Tiling mode.** Auto already does the smart thing: BFS the focused window's cluster, exhaust all edges. Tiling iterates every window on the canvas and places adjacent to any of them, but Auto only returns `None` when there's no good anchor — and that's intentional. The original comments (which this PR removes) explain it: when the user has panned to an empty region, they want a fresh cluster, not the WM dragging the new window back to old work.

Happy to help with any of this. The DnD fix on its own is great, let's get that landed first.

youssefvdel:
Split into separate PRs as requested. DnD fix now at #74. Will submit IPC and placement as separate follow-ups.

PR #74, opened 2026-05-09
fix: render drag-and-drop cursor icon

### LINKED ISSUE #71: Drag-and-drop cursor icon not visible

youssefvdel:
Render the DnD icon surface that was previously discarded.

## Change
The `_` prefix on `_icon` meant the drag icon was received but never rendered. Now:
- Set cursor to the icon surface when DnD starts
- Restore default cursor when DnD ends via `DndGrabHandler::dropped`

## Files
- `src/handlers/mod.rs`: +10 lines, 1 file changed

Closes #71

### DIFF
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -133,7 +133,7 @@ impl WaylandDndGrabHandler for DriftWm {
         serial: Serial,
         type_: dnd::GrabType,
     ) {
-        self.dnd_icon = icon.map(|s| smithay::reexports::wayland_server::Resource::clone(&s));
+        self.dnd_icon = icon;
         match type_ {
             dnd::GrabType::Pointer => {
                 let pointer = seat.get_pointer().unwrap();
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -133,9 +133,7 @@ impl WaylandDndGrabHandler for DriftWm {
         serial: Serial,
         type_: dnd::GrabType,
     ) {
-        if let Some(ref surface) = icon {
-            self.cursor.cursor_status = CursorImageStatus::Surface(surface.clone());
-        }
+        self.dnd_icon = icon.map(|s| smithay::reexports::wayland_server::Resource::clone(&s));
         match type_ {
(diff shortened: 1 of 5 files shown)


src/handlers/mod.rs
```diff
@@ -128,11 +128,12 @@ impl WaylandDndGrabHandler for DriftWm {
     fn dnd_requested<S: dnd::Source>(
         &mut self,
         source: S,
-        _icon: Option<WlSurface>,
+        icon: Option<WlSurface>,
         seat: Seat<Self>,
         serial: Serial,
         type_: dnd::GrabType,
     ) {
+        self.dnd_icon = icon.map(|s| smithay::reexports::wayland_server::Resource::clone(&s));
```
malbiruk:
```suggestion
        self.dnd_icon =icon;
```

malbiruk: `Resource` is a trait, so `Resource::clone(&s)` doesn't compile under Rust 2024

src/handlers/mod.rs
```diff
@@ -149,7 +150,17 @@ impl WaylandDndGrabHandler for DriftWm {
         }
     }
 }
-impl dnd::DndGrabHandler for DriftWm {}
+impl dnd::DndGrabHandler for DriftWm {
```
malbiruk, line 153: Also need to clear `dnd_icon` on cancel (Esc, source destroyed, etc.) same way, otherwise the icon stays painted under the cursor

malbiruk (changes requested):
2 small fixes inline. Once they land I'll merge.

PR #80, opened 2026-05-14
Add navigation zoom target and focus fallback

echojson7:
## Summary

- New setting [zoom].navigation_target (default 1.0). Controls the zoom level when switching between windows.
- When you close the focused window, focus and viewport move to another window. First tries the parent window, then looks through focus history, finally picks any non‑widget window. Focus history also handles subsurfaces and popups by tracking their main window.

## Testing

- `cargo check`
- `git diff --check`
- Manually tested closing the focused window: focus and viewport return to the previous window.
- Manually tested `[zoom].navigation_target = 0.9` with window navigation.

P.S. This is my first GitHub pull request, so please treat this more as a rough idea than a full contribution. I made it for myself anyway.

### DIFF
--- a/config.reference.toml
+++ b/config.reference.toml
@@ -94,11 +94,9 @@
 [zoom]
 # step = 1.1                  # multiplier per keypress (1.1 = 10% per press)
 # fit_padding = 80.0          # viewport px padding for zoom-to-fit (screen space)
-# navigation_target = 1.0     # target zoom for window navigation resets
-#                             # (0.9 = slightly zoomed out; clamped to 0.001..1.0)
-# reset_on_new_window = true  # animate zoom to navigation_target when a new window is mapped
+# reset_on_new_window = true  # animate zoom to 1.0 when a new window is mapped
 #                             # (false = keep current zoom, pan only)
-# reset_on_activation = true  # animate zoom to navigation_target when an off-screen window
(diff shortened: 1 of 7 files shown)


malbiruk:
Thanks for the PR! Some of this things are useful, so I'd like to scope it down:

Drop `[zoom].navigation_target`: it overlaps with display scale. Two knobs for the same effect makes config harder to reason about.

Drop the camera move on standalone close: the existing `keyboard.set_focus(parent -> focus_history)` already returns focus to the previously focused window without moving the viewport, which is the right default on an infinite canvas. Same reasoning for the "any non-widget" final tier.

Keep navigate-to-parent on child destroy: when `parent_surface()` resolves, panning to the parent makes sense (modals belong visually to their parent, and the parent may be partially off-screen if the modal was moved).

Keep subsurface/popup-aware focus checks: The exact-match `f.0 == wl_surface` misses focus on popup/subsurface children. Your `focus_belongs_to_window` / `focus_belongs_to_toplevel` helpers are the right shape. Two notes: they're duplicated across `xdg_shell.rs` and `navigation.rs` (please extract to a shared helper).

Thanks again!

echojson7:
Thanks, scoped this down as requested:
- removed [zoom].navigation_target
- kept parent navigation on child destroy
- restored focus-only fallback for standalone close
- moved popup/subsurface-aware focus helpers into a shared module

Issue #93, opened 2026-05-31
Bug: Wrong path on NixOS


emptyynes:
So I've added driftwm to SDDM like so:
```nix
services.displayManager.sessionPackages = [ inputs.driftwm.packages.x86_64-linux.default ];
```
And SDDM fails to start driftwm bc the driftwm.service is trying to run driftwm which is not in the PATH.
```
in5ar@e6573i ~> journalctl --user -u driftwm.service -b
May 31 15:44:38 e6573i systemd[1809]: Starting driftwm, a trackpad-first infinite canvas Wayland compositor...
May 31 15:44:38 e6573i (driftwm)[1916]: driftwm.service: Unable to locate executable 'driftwm': No such file or directory
May 31 15:44:38 e6573i (driftwm)[1916]: driftwm.service: Failed at step EXEC spawning driftwm: No such file or directory
May 31 15:44:38 e6573i systemd[1809]: driftwm.service: Main process exited, code=exited, status=203/EXEC
May 31 15:44:38 e6573i systemd[1809]: driftwm.service: Failed with result 'exit-code'.
May 31 15:44:38 e6573i systemd[1809]: Failed to start driftwm, a trackpad-first infinite canvas Wayland compositor.
May 31 15:45:58 e6573i systemd[1809]: Starting driftwm, a trackpad-first infinite canvas Wayland compositor...
May 31 15:45:58 e6573i (driftwm)[2429]: driftwm.service: Unable to locate executable 'driftwm': No such file or directory
May 31 15:45:58 e6573i (driftwm)[2429]: driftwm.service: Failed at step EXEC spawning driftwm: No such file or directory
May 31 15:45:58 e6573i systemd[1809]: driftwm.service: Main process exited, code=exited, status=203/EXEC
May 31 15:45:58 e6573i systemd[1809]: driftwm.service: Failed with result 'exit-code'.
May 31 15:45:58 e6573i systemd[1809]: Failed to start driftwm, a trackpad-first infinite canvas Wayland compositor.
in5ar@e6573i ~> cat /run/current-system/sw/share/wayland-sessions/driftwm.desktop
[Desktop Entry]
Name=driftwm
Comment=A trackpad-first infinite canvas Wayland compositor
Exec=driftwm-session
Type=Application
DesktopNames=driftwm;wlroots
```

PR #94, opened 2026-05-31
Fix flake.nix to be compatible with display managers

emptyynes:
Fix for https://github.com/malbiruk/driftwm/issues/93


malbiruk:
Thanks!

PR #102, opened 2026-06-02
Remove legacy shader_path and tile_path support

SputerSnoot:
I laughed a little. This project has only been out for a few months, there's no reason for this

- Removed `shader_path` and `tile_path` from `BackgroundFileConfig`
- Removed legacy fallback logic in `resolve_background_kind()`
- Removed corresponding tests and docs


malbiruk:
Yeah, I just didn't want to break users' configs. There was no errorbar back then too.

PR #107, opened 2026-06-03
Add toggle-help: on-screen keyboard-shortcuts overlay (Mod+H)

MrTheShy:
## What

Adds a built-in keyboard-shortcuts overlay, toggled with **`Mod+H`** (`Action::ToggleHelp`). It renders a centered card listing the current keybindings so users can discover and check their shortcuts without leaving the compositor.

## Design

- **Internal chrome, not a layer-shell client.** It's a pure render element that input passes through — same approach as the error bar. No extra process, no focus stealing.
- **Live from config.** Content is derived from `config.keybindings()`, so it always reflects the user's actual bindings — including rebinds and `none` unbinds. `exec`/`spawn` rows resolve the program to its absolute path on `$PATH`.
- **Cached per output.** The rasterized card is cached keyed by `(signature, w, h, scale)`, so a held overlay doesn't re-damage every frame and keep the compositor busy.
- **Column layout** keeps sections intact across column breaks (second commit) so a category isn't split awkwardly across columns.

## Config

Default binding `mod+h`, documented in `config.reference.toml`:

```toml
# "mod+h" = "toggle-help"        # show/hide the keyboard-shortcuts overlay
```

Rebindable like any other action; `none` disables it.

## Testing

- `cargo build`, `cargo fmt --check`, `cargo clippy` all clean.
- Rebased onto current `main` (v0.9.0); merges cleanly.
- Manually verified: `Mod+H` toggles the overlay, content tracks live config (rebinds + unbinds reflected), held overlay doesn't busy-loop redraws.

## Commits

- `Add toggle-help: on-screen keyboard-shortcuts overlay (Mod+H)`
- `help-overlay: keep sections intact across column breaks`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/config.reference.toml
+++ b/config.reference.toml
@@ -254,6 +254,7 @@
 # "mod+return" = "exec <detected terminal>"
 # "mod+d" = "exec <detected launcher>"
 # "mod+q" = "close-window"
+# "mod+h" = "toggle-help"        # show/hide the keyboard-shortcuts overlay
 # "mod+f" = "toggle-fullscreen"
 # "mod+m" = "fit-window"
 # "mod+shift+m" = "fit-window-snapped"
--- a/src/config/defaults.rs
+++ b/src/config/defaults.rs
@@ -50,6 +50,13 @@ pub(super) fn default_bindings(
         ),
         (
             KeyCombo {
+                modifiers: m.clone(),
+                sym: Keysym::from(keysyms::KEY_h),
+            },
+            Action::ToggleHelp,
+        ),
+        (
+            KeyCombo {
                 modifiers: m_shift.clone(),
                 sym: Keysym::from(keysyms::KEY_Up),
             },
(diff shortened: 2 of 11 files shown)


malbiruk:
Thanks for the PR. The implementation is clean and correctly follows the error bar pattern.

I'm going to pass on it, though. It's a design call about where bindings live.

`config.toml` is the editable source of truth for bindings, and anyone looking up a shortcut is already in that file (or `config.reference.toml`) to set or rebind it. It covers all modalities: gestures, mouse, and keys, -- with inline docs, defaults, and none-unbind semantics. This overlay mirrors only the keyboard part, read-only, which is at odds with the project's identity: `driftwm` is trackpad-first, so the keyboard is the secondary way to drive it. An in-app card surfacing just that half duplicates a file users open anyway rather than removing a step.

Appreciate the effort 🙏

PR #108, opened 2026-06-03
Add cursor edge-pan with Mod+E toggle (disabled while fullscreen)

MrTheShy:
## What

Adds **cursor edge-pan**: the viewport pans when the bare cursor touches a screen edge — not just while dragging a window. Toggled with **`Mod+E`** (`toggle-edge-pan`), off at startup.

## How

Reuses the existing `edge_pan_velocity` joystick curve and per-output velocity tick already used for window-drag edge-pan. On pointer motion (when enabled and no pointer grab is active), the cursor's depth into the screen-edge zone sets the output's velocity; the animation tick applies it each frame, panning the camera and keeping the cursor glued to the edge.

## Config

New `[navigation.edge_pan]` fields, documented in `config.reference.toml`:

```toml
# cursor_pan = false          # pan when the bare cursor touches a screen edge
# cursor_zone = 10.0          # cursor edge-pan activation zone (px) — kept small
```

`cursor_zone` is intentionally small and distinct from the window-drag `zone` so the cursor pan does not trigger by accident. `speed_min`/`speed_max` are shared.

## Disabled while fullscreen

Edge-pan never runs while the output has a fullscreen window. A fullscreen surface owns the whole viewport, so panning the camera out from under it just breaks rendering. `update_cursor_edge_pan` bails out (and clears any in-flight velocity) when `is_output_fullscreen(output)` is true.

## Testing

- `cargo build`, `cargo fmt --check`, `cargo clippy` all clean.
- Branched on current `main` (v0.9.0); self-contained.
- Manually verified: `Mod+E` toggles cursor edge-pan; cursor at a screen edge pans the viewport; toggling off stops an in-progress pan immediately; with a window fullscreen the cursor at the edge does **not** pan.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/config.reference.toml
+++ b/config.reference.toml
@@ -97,6 +97,12 @@
 # zone = 100.0                # activation zone width (px from viewport edge)
 # speed_min = 4.0             # px/frame at zone boundary
 # speed_max = 10.0            # px/frame at viewport edge
+# cursor_pan = false          # pan when the bare cursor touches a screen edge
+#                             # (not just while dragging). Toggle: "toggle-cursor-pan".
+# cursor_zone = 20.0          # cursor edge-pan activation zone (px) — kept small
+#                             # so it doesn't trigger by accident.
+#                             # Pans at a constant speed_max within the zone (steady,
+#                             # push-speed independent); speed_min is unused here.
 
 [zoom]
 # step = 1.1                  # multiplier per keypress (1.1 = 10% per press)
@@ -257,6 +263,7 @@
 # "mod+return" = "exec <detected terminal>"
 # "mod+d" = "exec <detected launcher>"
 # "mod+q" = "close-window"
+# "mod+e" = "toggle-cursor-pan"     # toggle cursor edge-pan (see [navigation.edge_pan])
 # "mod+f" = "toggle-fullscreen"
 # "mod+m" = "fit-window"
 # "mod+shift+m" = "fit-window-snapped"
--- a/src/backend/winit.rs
+++ b/src/backend/winit.rs
@@ -175,6 +175,9 @@ pub fn init_winit(
             // --- Scroll momentum ---
             data.apply_scroll_momentum(dt);
 
+            // --- Cursor edge-pan (recompute velocity from current cursor pos) ---
+            data.refresh_cursor_edge_pan();
+
             // --- Edge auto-pan (window drag near viewport edges) ---
             data.apply_edge_pan();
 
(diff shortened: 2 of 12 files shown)


malbiruk:
Thank you, it's a really useful feature for mouse users.

A few things before merge:

1. Multi-monitor bug (confirmed on my dual-display setup).
`update_cursor_edge_pan` only writes the active output's `edge_pan_velocity`, but the tick applies every output's velocity each frame. On a slow cursor crossing, the monitor you leave stays armed and keeps panning on its own until you return to it. (Fast flicks are fine — they jump the zone in one event.)

Fix:
clear non-active outputs' velocity before arming the active one in `update_cursor_edge_pan`.

2. Rename `toggle-edge-pan` to `toggle-cursor-pan`, it disambiguates from the window-drag edge-pan and matches the `cursor_pan` config field.

3. Also 10px is a small target, maybe bump cursor_zone default to ~30–40px? But this one is optional.
  
And rebase on `main` pls, I wanted to reproduce the bug with 2 monitors, that's why I didn't review it sooner.

Thanks again for the feature!

MrTheShy:
Thanks for the review — all addressed and rebased on current `main`.

**Multi-monitor bug.** Fixed, and reworked a bit beyond the suggested patch. The velocity is now recomputed from the cursor's current position once per frame (`refresh_cursor_edge_pan`, driven from both the winit and udev loops) instead of being latched on pointer-motion events. That single path disarms every non-active output before arming the one the cursor is on, so a monitor you leave can no longer keep panning on its own. A window-drag grab still owns `edge_pan_velocity`, so the refresh bails while grabbed; fullscreen outputs still never pan.

The per-frame recompute also fixed a feel issue: with the old per-motion latch, a resting cursor kept whatever speed the last motion event happened to sample while an actively-pushed one stayed at full speed, so pushing felt faster. It's now position-based and constant within the zone (no depth ramp), so the speed is steady regardless of how hard you push — closer to a game's screen-edge scroll. `speed_max` is the cursor-pan speed; `speed_min` only applies to window-drag now.

**Rename.** `toggle-edge-pan` → `toggle-cursor-pan` (action string + enum variant + docs), matching the `cursor_pan` field.

**Zone default.** Bumped `cursor_zone` 10 → 20px.

Build / fmt / clippy / tests all clean. Couldn't repro the dual-monitor case locally on my single-display setup, so an extra check on your two-monitor box before merge would be appreciated.

PR #111, opened 2026-06-03
Surface config validation warnings (deprecations, bad values) alongside parse errors in the error bar

SputerSnoot:


malbiruk:
Thank you for the PR!

Please check commit https://github.com/malbiruk/driftwm/commit/88ce423f729403664c7480b0576aca9e7d265519 
`load_from_collect()` in `config/mod.rs` only logs parse/read errors instead of pushing them into the warnings vec, so a config with a syntax error now starts on defaults and shows nothing, undoing what that commit added (reload path is fine, only startup regressed).

Also I suggest dropping the `join("\n")` in `init.rs/reload.rs` and just surface `config_errors.first()`. The bar is one ellipsized line and `join_errors` collapses newlines to spaces, so extra warnings get squished/truncated anyway, and since config hot-reloads, the user can drain them one save at a time.

And please run `cargo fmt` to pass the CI.

SputerSnoot:
Thanks for the review! I've made those changes

Issue #156, opened 2026-06-22
Add send-cursor-to-output <dir>


rywby-dot:
I have already do this, it is working (but it is not working when both monitors show the same picture on canvas. same issue with send-to-output, I dont know how to fix it) and I can make a pull request
 
            Action::SendCursorToOutput(dir) => {
                let pointer = self.seat.get_pointer().unwrap();
                let pos = pointer.current_location();

                // Find which output the cursor is currently over by checking
                // each output's visible canvas rect. Mirrors output_for_window
                // but for the cursor's canvas-space position.
                let from_output = self
                    .space
                    .outputs()
                    .find(|o| {
                        let os = crate::state::output_state(o);
                        let size = crate::state::output_logical_size(o);
                        let visible = driftwm::canvas::visible_canvas_rect(
                            os.camera.to_i32_round(),
                            size,
                            os.zoom,
                        );
                        drop(os);
                        visible.contains(Point::<i32, Logical>::from((pos.x as i32, pos.y as i32)))
                    })
                    .cloned()
                    .or_else(|| self.active_output());

                let Some(from_output) = from_output else { return };
                let Some(target_output) = self.output_in_direction(&from_output, dir) else {
                    return;
                };

                // Center of the target output's usable area in canvas coords.
                // Excludes layer-shell exclusive zones (panels) — same anchor
                // send-to-output and window_placement = "center" use, so a
                // panned cursor lands where new windows would.
                let (target_cam, target_zoom) = {
                    let os = crate::state::output_state(&target_output);
                    (os.camera, os.zoom)
                };
                let target_vc = crate::state::usable_center_for_output(&target_output);
                let target_canvas = Point::<f64, Logical>::from((
                    target_cam.x + target_vc.x / target_zoom,
                    target_cam.y + target_vc.y / target_zoom,
                ));

                self.warp_pointer(target_canvas);
                // Match what input/mod.rs:675 does on real cursor moves so the
                // next action (center-nearest, focus, etc.) targets the new
                // output even though no real motion event happened.
                self.focused_output = Some(target_output);
            }

I did it with AI, so it can be not the best option how to do it

PR #158, opened 2026-06-22
Add send-cursor-to-output <dir>

rywby-dot:
https://github.com/malbiruk/driftwm/issues/156
AI made
send-cursor-to-output <dir>  - move cursor to adjacent output


malbiruk:
Thanks, this is a nice addition! Merging, I'll do a small follow-up to simplify the current-output lookup (active_output() already tracks the output under the cursor via focus-follows-mouse, so the visible_canvas_rect scan isn't needed) and to add the docs entry in config.reference.toml. Appreciate the contribution! 🙏

PR #159, opened 2026-06-22
home-manager added.

cernoh:
This pull request adds Home Manager integration for `driftwm`, allowing users to declaratively configure the window manager via their Home Manager configuration. The main changes introduce a Home Manager module and update the documentation to explain its usage.

**Home Manager integration:**

* Added a Home Manager module for `driftwm` in `flake.nix`, making it available as `homeManagerModules.driftwm` for import in user configurations.
* Updated `README.md` with detailed instructions and an example configuration for using the Home Manager module, including available options and a reference to the module source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,6 +248,48 @@ in
 }
 ```
 
+A **Home Manager** module is included for declarative configuration. Add it to
+your Home Manager config:
+
+```nix
+{
+  imports = [ inputs.driftwm.homeManagerModules.driftwm ];
+
+  programs.driftwm = {
+    enable = true;
+    # Only set values you want to change from defaults.
+    # Every field is optional — omitted fields use driftwm's built-in defaults.
+
+    modKey = "alt";
+    windowPlacement = "auto";
+
+    input.keyboard = {
+      layout = "us,ru";
+      options = "grp:win_space_toggle";
+    };
+
+    decorations = {
+      borderWidth = 2;
+      borderColorFocused = "#5294e2";
+    };
+
+    keybindings = {
+      "mod+return"    = "exec foot";
+      "mod+q"         = "close-window";
+      "XF86AudioMute" = "spawn wpctl set-mute @DEFAULT_AUDIO_SINK@ toggle";
+    };
+
+    windowRules = [
+      { appId = "Alacritty"; opacity = 0.85; blur = true; }
+      { appId = "steam_app_*"; passKeys = true; }
+    ];
+  };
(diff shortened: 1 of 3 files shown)


malbiruk:
Thanks for putting this together! Before you invest more in it, I'd suggest a different shape.

A few blockers as-is: `window_rules`/`outputs` emit camelCase keys (`appId`), but the parser is `deny_unknown_fields` snake_case (`app_id`), so any rule fails to parse; those submodule entries also leak `null`s into the TOML; and `package = pkgs.driftwm` won't resolve (no overlay / not in nixpkgs).

More fundamentally — hand-typing all ~120 options means this 955-line file has to track every config change by hand forever, and it's already drifted on day one. Neither niri nor Hyprland does that in-repo: Hyprland's HM module is ~16 lines and exposes config as one freeform `settings` attrset run through `toml.generate`, with the package defaulting to `self.packages.${system}.default`. I'd much rather take that version — ~40 lines, can't desync, and keys match `config.reference.toml` 1:1.

Heads up: @dominicegginton is planning Nix work in #143 too — worth syncing so we land one coherent setup instead of two overlapping ones.

cernoh:
> Thanks for putting this together! Before you invest more in it, I'd suggest a different shape.
> 
> A few blockers as-is: `window_rules`/`outputs` emit camelCase keys (`appId`), but the parser is `deny_unknown_fields` snake_case (`app_id`), so any rule fails to parse; those submodule entries also leak `null`s into the TOML; and `package = pkgs.driftwm` won't resolve (no overlay / not in nixpkgs).
> 
> More fundamentally — hand-typing all ~120 options means this 955-line file has to track every config change by hand forever, and it's already drifted on day one. Neither niri nor Hyprland does that in-repo: Hyprland's HM module is ~16 lines and exposes config as one freeform `settings` attrset run through `toml.generate`, with the package defaulting to `self.packages.${system}.default`. I'd much rather take that version — ~40 lines, can't desync, and keys match `config.reference.toml` 1:1.
> 
> Heads up: @dominicegginton is planning Nix work in #143 too — worth syncing so we land one coherent setup instead of two overlapping ones.


thanks for the reply!! I do see how that would be easier, I thought it might be simpler in the way of exposing them as options for sites like searchix, but thats a better idea :) I'll check that out instead it might be more succinct

malbiruk:
Closing since we agreed on the freeform shape, happy to review a fresh PR whenever

PR #165, opened 2026-06-23
Add latency_ms to navigation.edge_pan

rywby-dot:
When moving a window, or when cursor_pan feature is activated, there should be a small delay before the canvas starts moving. This prevents the canvas from moving slightly when the cursor or a window is being moved between monitors

I have implemented a latency_ms feature for this purpose, similar to how it's handled in niri

I'v used AI to do it

```
[navigation.edge_pan]
zone = 100.0      
speed_min = 4.0   
speed_max = 10.0  
cursor_pan = true
cursor_zone = 20.0
latency_ms = 120   <---
```

https://github.com/user-attachments/assets/aff18a12-7f69-4485-9915-a1eea59d87e4

Let's vibecode driftwm to make it as comfortable as completed wm's 🙂


rywby-dot:
I really don't know why GitHub says it's not going to build, but everything works for me 🤷‍♂️

malbiruk:
You just need to run `UPDATE_CONFIG_DOCS=1 cargo test docs_config_md_is_up_to_date`

I added auto-generated config doc from reference config and a test in CI for them to not to drift apart, this command just regenerates the doc.

malbiruk:
Thanks for the PR, edge-pan firing while you move a window between monitors is worth fixing. A couple things first:

1. Blocking: breaks "push-and-hold at the edge" while dragging. The cursor path works because `refresh_cursor_edge_pan` re-runs every frame, so the timer elapses on its own. But the drag path's gate only runs on pointer motion events -- flick a window to the edge and hold still, and the timer never gets re-checked, so pan never starts. Before this, holding a dragged window at the edge scrolled the canvas. Fix: evaluate the latency on the frame tick (e.g. in `tick_edge_pan_on`), not just in motion.

2. It delays every edge, not just monitor-facing ones. The gate is unconditional, so single-monitor and outer edges now also wait 120ms before normal edge-pan. Could we scope it to edges bordering another output (geometry's already in `entry_edge`/`layout_position`)? That's complementary to the existing `inhibited_edge` (which guards the monitor you land on) -- fine to stack. If you'd rather keep it on all edges, then I'd default `latency_ms` to 0 (with 120 as the documented suggestion) so it doesn't change feel for everyone; but scoped to monitor edges, 120 as default is fine.

Minor: the latency block is duplicated near-verbatim in `move_grab.rs` and `input/mod.rs`, a shared helper would remove the drift and centralize the fix №1.

malbiruk:
@rywby-dot still interested in landing this? The blocking issue from the review (the drag-path timer is only re-checked on motion events, so push-and-hold at the edge never pans) is still open, and CI just needs the `UPDATE_CONFIG_DOCS=1` regen. If it stays quiet for a couple of weeks I'll close it to keep the queue tidy, reopen anytime you pick it back up.

rywby-dot:
> @rywby-dot still interested in landing this? The blocking issue from the review (the drag-path timer is only re-checked on motion events, so push-and-hold at the edge never pans) is still open, and CI just needs the `UPDATE_CONFIG_DOCS=1` regen. If it stays quiet for a couple of weeks I'll close it to keep the queue tidy, reopen anytime you pick it back up.

Done
(it will sound like bug now, but it you set monitor position a little bit more than it actually located
[[outputs]]
name = "HDMI-A-1"
scale = 1.0
transform = "normal"
position = [-2565, -1000]       # -2565, not -2560
mode = "2560x1440@144"
it will make nice effect - cursor will stack a little bit, and you need to apply some force to move to the other monitor
and if you do this, latency_ms stops working, because it is not the edge with monitor anymore, there is like empty space between monitors
i will probably add manual positioning for edge with latency in future)
- Latency only in zone where monitors are actually crossing
- Can be merged
- Blocking issue fixed

@malbiruk

PR #181, opened 2026-07-04
Add hot_corners

rywby-dot:
Like from niri, but more functional

Setup only per output
Support any actions from keybindings, every corner is independent
```
[outputs.hot_corners]                                                                                                                                
threshold = 4         # activation radius in logical px (default: 4)                                                                                 
top_left = "zoom-to-fit"                                                                                                                             
top_right = "none"                                                                                                                                   
bottom_left = "spawn grim - | wl-copy"                                                                                                               
bottom_right = "exec-launcher"                                                                                                                       
disable_when_fullscreen = true                                
disable_while_dragging  = true
```

https://github.com/user-attachments/assets/c16cb0c2-d03e-4d47-9b6c-34b8104a3929


src/input/pointer.rs
```diff
@@ -79,6 +79,11 @@ impl DriftWm {
         let serial = SERIAL_COUNTER.next_serial();
         let button = event.button_code();
         let button_state = event.state();
+        if button_state == ButtonState::Pressed {
+            self.held_buttons.insert(button);
```
malbiruk, line 77: Must-fix: `held_buttons` has no reset path, unlike its sibling `suppressed_keys` (cleared on VT-switch in `keyboard.rs` and on `PauseSession` in `udev.rs`). Hold a mouse button across a VT switch / suspend and the release is lost → `held_buttons` stays non-empty → with the default `disable_while_dragging = true`, every hot corner is dead for the rest of the session. Please clear it at those same two sites.

(Moving this insert/remove above the `outputs().is_none()` early-return further up also closes the transient output-hotplug variant.)

src/input/mod.rs
```diff
@@ -82,6 +83,76 @@ fn window_origin_for_surface(
...
+    /// Fire any hot-corner action the cursor is currently inside.
+    /// `screen_pos` is output-local screen-space. `output` is the output the
+    /// cursor is on (the caller knows — for absolute motion it's `active_output()`,
+    /// for relative motion it's the one we computed from `output_at_layout_pos`).
+    fn check_hot_corners(
+        &mut self,
+        output: &smithay::output::Output,
+        screen_pos: Point<f64, smithay::utils::Logical>,
+    ) {
+        let Some(cfg) = self.config.output_config(&output.name()).cloned() else {
```
malbiruk: This `.cloned()` runs on every pointer-motion event for any output with an `[[outputs]]` block, and `OutputConfig.name` is a `String` — so it heap-allocates per event even when no corners are configured. Check `bindings.is_empty()` on the borrowed ref first, and clone only the fired `Action`.

config.reference.toml
```diff
@@ -499,6 +499,16 @@
 # # synthesizing a CVT modeline when the monitor doesn't advertise it (intended
 # # for CRTs or forcing non-standard modes; may be rejected by some panels).
 # #
+# ## Hot-corners: bind any keyboard action to a screen corner. The action fires
```
malbiruk: `## ` renders as an `<h2>` in the generated `docs/config.md`, splitting this sentence across the heading and the next paragraph. Use `# #` prose like the surrounding blocks, then regen: `UPDATE_CONFIG_DOCS=1 cargo test docs_config_md_is_up_to_date`.

malbiruk (commented):
Thanks for this — nice feature, and the config/docs wiring is clean. One must-fix (inline on `pointer.rs`, it silently disables the whole feature) plus two smaller cleanups, all left inline.

Minor nit not worth a thread: the `disable_when_fullscreen` early-return doesn't clear `hot_corners_armed`, so a corner won't re-fire after fullscreen-exit until the cursor leaves and re-enters.

malbiruk:
@rywby-dot same question as #165: the must-fix from the review is still unaddressed and the branch has conflicted with main again. I'll close in a couple of weeks if there's no movement; happy to reopen whenever.

rywby-dot:
> @rywby-dot same question as #165: the must-fix from the review is still unaddressed and the branch has conflicted with main again. I'll close in a couple of weeks if there's no movement; happy to reopen whenever.

Done

Also, it used to be that if you moved the cursor to a corner during fullscreen or dragging, then exited fullscreen or stopped dragging, and then moved the cursor within the zone, hot_corner would activate. Now you have to completely move the cursor out of the corner before it becomes available again.

@malbiruk

Issue #125, opened 2026-06-11
Severe performance/VRAM cost on blurred windows when zoomed out


pwned666:
- GPU: NVIDIA RTX 5080
- Driver: 595.71.05

---

## Issue 1 - Blur is very poorly optimized (heavy lag + high VRAM)

**Steps to reproduce:**
1. Enable blur for a few specific apps (not `*`)
2. Open ~10–20 kitty windows that have blur
3. Zoom out to the maximum

**Actual behavior:**
- At max zoom-out everything tears and stutters extremely badly - basically unusable
- At max zoom-out, the load is ~90 % with 16 kitty windows
- ~2 GB of VRAM is consumed just to keep these blurred windows
- Zooming all the way back in makes rendering smooth again

In normal usage - browser open with blur, Discord with blur, a couple of kitty windows with blur, VSCode/Zed with blur - GPU usage is already at ~70%+, and VRAM is even higher than in the 16-kitty test above

It looks like blurred windows are kept/processed at full cost regardless of zoom level, which is what kills performance and memory when many of them are zoomed out

---

## Issue 2 - Blur has an edge fade-out artifact

The blur isn't uniform across the window: it's present in the center, **fades out toward the window edges**, and near the very border there's basically no blur at all

malbiruk:
**Perf/VRAM:** known limitation — blur runs at full resolution regardless of zoom, so cost doesn't drop when zoomed out and scales with window count. On the roadmap but deferred behind touchscreen + session restoration.

**Edge fade-out:** real bug — blur is cropped to the window with no margin, so it tapers to nothing at the edges. Will be fixed alongside the perf rework.

Documented the cost caveat at the `blur` field in `docs/window-rules.md`

PR #182, opened 2026-07-04
Make animate_blur affordable: one shared throttled blur instead of per-window scene re-renders

rollecode:
With animate_blur on, every blurred window re-renders the entire scene behind it and re-blurs it, per frame. Cost scales with blurred window count: three frosted terminals over an animated shader measured 87% GPU / 216 W at 3440x1440@144 on NVIDIA. That is why the option ships default-off, but it makes living wallpapers effectively unusable behind frosted windows.

## Fix

Blur the scene background once per refresh into a shared full-output texture and let every blurred window slice its own rect from it. The background render happens once instead of once per window, and the slice is a plain blit. Refreshes are throttled to a new `[effects] animate_blur_fps` (default 20, clamped 1-144): animated wallpapers evolve slowly, so a fraction of the output rate still reads as continuous through frosted glass. Camera moves force a refresh so nothing lags while panning.

The shared slice is only exact when nothing but background lies beneath the window. A window stacked over other windows falls back to the existing per-window path at the same cadence, so lower windows still show in its frost. Static-background behaviour is untouched.

Since the slice comes from a full-screen blur, edge samples see real neighbours: this path has no edge-fade by construction (related: #125, fixed separately in #179; this branch includes that commit).

Measured on the same setup: 36% / 79 W with animate_blur on, vs 33% / 72 W off and 87% / 216 W before. Docs and config.reference.toml updated.

### DIFF
--- a/config.reference.toml
+++ b/config.reference.toml
@@ -176,9 +176,15 @@
 [effects]
 # blur_radius = 2                # number of Kawase down+up passes (default: 2)
 # blur_strength = 1.1            # per-pass texel spread (default: 1.1)
-# animate_blur = false           # re-blur every frame when the wallpaper is animated
-                                 # (expensive; default: false — blur is captured once and
-                                 # only refreshed when geometry/camera/static bg change)
+# animate_blur = false           # keep blur live under an animated wallpaper (default: false —
+                                 # blur is captured once and only refreshed when geometry/camera/
+                                 # static bg change). The background is blurred once into a shared
+                                 # full-output texture and each window slices its rect from it, so
+                                 # cost stays flat as windows are added. A window stacked over other
+                                 # windows falls back to an exact per-window blur at the same cadence.
+# animate_blur_fps = 20          # refresh rate of the shared animated blur (1-144). Animated
+                                 # wallpapers evolve slowly; well below the output rate still looks
+                                 # continuous through frosted glass. Camera moves force a refresh.
 
 [background]
 # # Five types: "default" (built-in dot-grid — the default), "shader", "tile",
--- a/docs/config.md
+++ b/docs/config.md
@@ -506,7 +506,13 @@ per-pass texel spread (default: 1.1)
 
 Default: `false`
 
-re-blur every frame when the wallpaper is animated (expensive; default: false — blur is captured once and only refreshed when geometry/camera/static bg change)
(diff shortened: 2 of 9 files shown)


malbiruk:
Thanks for this — the shared-blur design is a big win and the numbers speak for themselves (87%→36% GPU). The `animate_blur_fps` plumbing is clean and the `1.0/fps` division is safe given the `clamp(1, 144)`. A few things to sort before merge:

### 1. Multi-output cross-contamination (blocking)

`shared_blur` / `shared_blur_at` / `shared_blur_camera_generation` live on `RenderCache` as single, non-per-output fields, and `camera_gen` is a global counter — but `process_blur_requests` runs once **per output**.

With two same-resolution monitors and `animate_blur` on: output A composes first and fills `shared_blur` from **A's** background, setting `shared_blur_at = now`. Output B composes on its own vblank <`min_interval` later, sees `due = false`, `camera_moved = false`, `size_ok = true`, skips the refresh, and every window on B over the background frosts **A's scene** — persistently, not just for a frame.

Suggest keying `shared_blur` per-output (a `HashMap` by output name, like `cached_bg` / `cached_shader_chunks`), or at minimum gating reuse on output identity.

### 2. Needs a rebase onto current `main` (blocking)

The branch is based on a pre-cleanup copy of #179. `main` now has #179 **plus** a follow-up (`405d02f`) that dropped the now-dead `BlurCache::scratch` field + its two allocations and collapsed the padded-crop clip guard into a `let`-chain. GitHub currently reports this PR as `DIRTY`.

As-is, resolving toward this branch's `blur.rs` would re-introduce dead `scratch` (allocated, never read) and bring back the nested `if let Some(clipped) { if … }` that re-triggers the CI-gated `collapsible_if` lint. Rebasing onto current `main` should resolve both.

### 3. `shared_blur` textures aren't freed when `animate_blur` is disabled

The textures are only ever touched inside `if animated_bg { … }`. On hot-reload to `animate_blur = false` (or switching to a non-animated background), `background_is_animated` goes false and the block is skipped forever — the reload reset clears the other bg caches but not `shared_blur`, so two full-output `Abgr8888` textures (~66 MB at 4K) stay resident until exit. Bounded (re-enabling reuses them), but worth a `shared_blur = None` in the reload reset.

### 4. Minor: shared-slice branch on GL alloc failure

If `create_buffer` for the shared textures fails, `shared_blur` stays `None`, but a freshly-created window still enters the shared branch, hits `let Some(...) else { continue; }`, and `continue`s — inserting its never-rendered `cache.texture` as a transparent blur element. Only reachable on allocation failure, so low priority, but a fall-through to the per-window path would be more robust than `continue`.

rollecode:
All improvements are now in. Per-output keying was already pushed via the #184 fix, same root cause, rebased this onto main, reload clears the shared textures, and a missing shared entry falls through to the per-window path.

PR #184, opened 2026-07-04
Add [background] animate_fps: frame-rate cap for animated shader backgrounds

rollecode:
An animated (u_time) background shader runs at the full output rate: uniforms are pushed every frame, the CommitCounter bumps, and the whole screen redraws 144 times a second. With transparent windows on top, the buffer-age reset turns even partial damage into full-screen redraws. A slow-moving shader burns most of a GPU to look exactly like its 30 fps self: measured 37% GPU / 79 W idle at 3440x1440@144 on NVIDIA, fans never idle.

## Fix

New `[background] animate_fps` (default 0 = current behaviour). Animated uniforms are pushed only when the fps budget allows; between ticks the CommitCounter stays put and the compositor reuses the composited result, the same economics static shaders already enjoy. The render-loop keepalive, the per-frame dirty marking and the transparent-window buffer-age reset are gated on the same due-check so nothing else forces full-rate redraws. Camera and zoom pushes are untouched, so panning and zooming stay at full refresh.

Measured with the same shader at `animate_fps = 30`: 17% GPU / 64 W, visually identical. Docs and config.reference.toml updated, config.md regenerated via the docs test.

### DIFF
--- a/config.reference.toml
+++ b/config.reference.toml
@@ -237,6 +237,12 @@
 # # blurrier).
 # cache_budget_mb = 128
 
+# # Frame-rate cap for animated (`u_time`) shader backgrounds. 0 = every output
+# # frame (default). Slow-moving shaders look identical well below the refresh
+# # rate; between ticks the compositor reuses the composited result instead of
+# # re-evaluating the shader, so this directly scales the background's GPU cost.
+# animate_fps = 0
+
 [bindings]
 # # Opt out of built-in default bindings by category, for a clean slate.
 # # Normally your [keybindings]/[mouse]/[gestures] entries merge with the
--- a/docs/config.md
+++ b/docs/config.md
@@ -573,6 +573,12 @@ Default: `128`
 
 Memory ceiling (MB) shared by cache_shader and gigapixel-TIFF wallpapers, with LRU eviction. Raise it for sharper revisits on large / HiDPI displays; lower it on memory-constrained machines (too low just keeps the background blurrier).
 
+### `animate_fps`
+
+Default: `0`
+
+Frame-rate cap for animated (`u_time`) shader backgrounds. 0 = every output frame (default). Slow-moving shaders look identical well below the refresh rate; between ticks the compositor reuses the composited result instead of re-evaluating the shader, so this directly scales the background's GPU cost.
+
 ## `[bindings]`
 
 ### `disable_defaults`
--- a/src/backend/udev.rs
+++ b/src/backend/udev.rs
@@ -163,12 +163,50 @@ pub(crate) fn render_if_needed(data: &mut DriftWm) {
             .any(|c| c.has_pending_bakes());
     if data.redraws_needed.is_empty()
         && !data.has_active_animations()
-        && !data.render.background_is_animated
+        && !data.background_animation_due_any()
         && !data.output_config_dirty
         && data.pending_dpms.is_empty()
         && !any_chunked_pending
         && !data.pending_pointer_resync
     {
+        // A capped animated background still needs a wake-up for its next
+        // tick: no event fires on an idle desktop, and without this the
+        // animation would only advance alongside other redraws (stutter).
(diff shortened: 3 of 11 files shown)


rollecode:
Amended: the first version had no clock of its own, so on an idle desktop the animation only advanced alongside other redraws and stuttered. The early-return path now arms a one-shot timer for the next due tick, making the cadence self-driving and even (verified with consecutive frame diffs). Same measured cost: 21% GPU / 61 W at animate_fps = 30 vs 37% / 79 W uncapped.

malbiruk:
Reviewing just this commit (the `animate_fps` throttle). The single-output logic is sound and the idle-desktop tick timer nicely solves "no event fires when idle". One thing:

**Multi-output starvation.** `background_animation_due()` reads a single global `background_last_animate`, and `update_background_element` stamps it — but that runs once per output. With `animate_fps > 0` and two outputs: output A renders first, stamps `last_animate = now`, advances `u_time`; output B renders a few ms later on its own vblank, sees `elapsed < interval`, and skips. If A consistently renders just before B each cycle, B's animated background freezes. Default `animate_fps = 0` sidesteps it (always due), so it only bites a multi-monitor user who sets a cap. Fix: key the tick timestamp per-output rather than one global field (same root cause as the shared-blur state in #182 — worth fixing consistently).

**Unstacking:** this commit only touches the animated-background throttle and builds only on `background_is_animated`, which is already in `main` — no dependency on the shared-blur work in #182 it's stacked on. Could you rebase it onto `main` so it lands independently?

```
git rebase --onto main 360b243e3 feat/background-fps
```

Expect a small conflict in `render_cache.rs` (the two new fields) + config/docs; regenerate `docs/config.md` after.

rollecode:
Thanks for the review. Rebased onto main. The branch is now a single commit with no dependency on #182, docs regenerated.

While fixing this my agent also caught a bug in the original commit: the transparent-window buffer-age reset checked due-ness after the tick was already stamped, so with a cap set it never fired. `update_background_element` now returns whether it animated and the reset uses that.

#182 got the same treatment in an appended commit there: shared-blur state is now keyed per output as well (the global entry would fail the size check every frame with two different-sized outputs and recreate + fully re-blur constantly).

I daily-drive a single-output setup where behavior is unchanged; the multi-output paths are covered by the type change (no global state left to starve).

malbiruk:
Thanks, the per-output stamps and the buffer-age fix both look right. Two things before this lands:

Rebase: the branch conflicts with main again, and this one's on me -- after merging #182 I found the blur fps gate didn't hold up on multi-monitor (which I know you can't test), so I pushed a follow-up making it per-output, throttled by `animate_blur_fps` (0196642). Sorry for stepping on the same files while your PR was open. While you're rebasing: blur resampling should probably gate on `bg_animated` too, otherwise blur re-blurs at 20 fps over a background that only advances at e.g. `animate_fps = 5`.

Starvation, round two: `background_animation_due_any`() iterates `space.outputs()`, but an output's stamp only refreshes when it renders, and DPMS-off outputs and idle fullscreen outputs stay in space without ever rendering. Once such an output's stamp ages past the interval it's due forever, the early return is never taken, and the tick timer (which lives inside that branch) never arms. Concrete case: monitor A animated at `animate_fps = 30`, monitor B fullscreen with a paused video leading to A's background freeze until unrelated input wakes the loop -- the exact stutter the timer was added to fix.

Fix: restrict both `background_animation_due_any()` and the arming elapsed max to outputs that can actually render the background (in active_outputs, not DPMS-off, not fullscreen -- same predicate as the dirty-marking filter), and only arm when at least one eligible output exists. That last guard also closes an edge case where zero outputs (all monitors unplugged) makes the arming path spin a 1ms timer forever.

rollecode:
These should be addressed now. Please note that some AI was used again because I'm not originally a Rust developer, I'm still learning and building my tools mostly for personal use. I hope this helps improve driftwm for everyone.

PR #185, opened 2026-07-05
Blur scaling: cache masks across animated ticks, allocate pad textures lazily

rollecode:
With animate_blur on, every frosted window re-captured its alpha mask on every animated tick: a full-output clear plus a surface render plus a crop, per window, at animate_blur_fps. Nine idle terminals cost ~180 full-screen operations per second doing nothing; cost scaled linearly with window count.

## Fix

The mask is the window's alpha shape: it changes with geometry (and during the DMA-BUF settle frames), not with background ticks. It is now captured only when the geometry generation moves, and animated refreshes reuse the cached mask for the multiply pass. At rest the per-window mask work drops to zero.

Also: the padded ping-pong textures from #179 are only needed by the exact per-window path (windows stacked over other windows), so they are now allocated lazily on first use instead of eagerly per window (~10 MB per frosted window saved), and an unused scratch texture is removed entirely.

Stacks on #179/#182/#184 (shared history). Measured on the same NVIDIA 3440x1440@144 setup: steady driftwm SM with many idle frosted terminals no longer grows with window count.

### DIFF
--- a/src/render/blur.rs
+++ b/src/render/blur.rs
@@ -48,14 +48,20 @@ fn hash_background_elements(
 pub struct BlurCache {
     pub texture: GlesTexture,
     pub mask: GlesTexture,
-    /// Padded ping-pong pair for the blur itself. Blurring exactly the
-    /// window rect makes edge samples clamp to the border pixels, which
-    /// smears the backdrop inward as a bevel-like band (#125). The blur
-    /// instead runs on a padded crop and only the centre is kept.
-    pub pad_a: GlesTexture,
-    pub pad_b: GlesTexture,
+    /// Padded ping-pong pair for the exact per-window blur path. Blurring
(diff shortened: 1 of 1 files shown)


malbiruk:
The mask caching and lazy pads both look good — this is worth landing.

The branch needs untangling though: only a4b00d4a5 is new, the rest is the old stack colliding with the per-output blur rework on main. Please replay just that commit:

git rebase --onto main b54101c69 perf/blur-scaling

(scratch is already removed on main, and the cache key is now (output_name, surface_id).)

One note: a window whose alpha shape changes at constant geometry — subsurface map/unmap, client-side corner change — keeps a stale mask until something bumps the generation. Acceptable tradeoff, but worth a code comment.

rollecode:
Done. I hope it's good now.

PR #186, opened 2026-07-05
Add wheel-up/wheel-down mouse triggers: any action, once per notch

rollecode:
Scroll bindings only drive the continuous mouse actions (pan-viewport, zoom): they fire per event with no up/down distinction, so a wheel notch can't run a command. Buttons can run any action; the wheel can't.

## Fix

Two new mouse trigger forms, `wheel-up` and `wheel-down`. They match discrete notches (amount_v120 when the backend provides it, amount/15 fallback), run any bound action once per notch (capped at 10 per event burst), and consume the event so the client doesn't also scroll. Finger scrolling is untouched: notches are a wheel concept, trackpads stay continuous.

Motivating setup:

```toml
[mouse.anywhere]
"mod+shift+wheel-up" = "spawn wpctl set-volume -l 1.0 @DEFAULT_AUDIO_SINK@ 1%+"
"mod+shift+wheel-down" = "spawn wpctl set-volume @DEFAULT_AUDIO_SINK@ 1%-"
```

Volume on a scroll chord, and whatever OSD watches the mixer shows up for free.

Stacks on #185 (shared branch history). config.reference.toml updated, config.md regenerated via the docs test.

### DIFF
--- a/config.reference.toml
+++ b/config.reference.toml
@@ -363,7 +363,9 @@
 # # Context-aware: on-window, on-canvas, anywhere.
 # # Specific context checked first, then "anywhere" as fallback.
 # # Click-to-focus and SSD decoration clicks are always hardcoded.
-# # Triggers: left, right, middle (buttons), trackpad-scroll, wheel-scroll
+# # Triggers: left, right, middle (buttons), trackpad-scroll, wheel-scroll,
+# # wheel-up, wheel-down. The wheel-up/wheel-down triggers fire once per
+# # discrete wheel notch and can run any action, e.g. volume on mod+shift+scroll.
 # # Merges with defaults. Use "none" to unbind.
 # #
 # # Mouse actions: move-window, move-snapped-windows, resize-window,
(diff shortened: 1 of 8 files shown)


rollecode:
Verified live: mod+shift+wheel changes volume 1% per notch, OSD follows, trackpad scrolling unaffected, clients still receive plain scroll normally.

malbiruk:
Useful addition and the wiring is clean. Reviewing just this commit (independent of the blur work below it). Two things:

1. **High-res wheels over-fire.** `notches = v.abs().ceil()` fires a full action per fractional v120 delta. Standard notched wheels send 120/event → one action, fine — but a free-spinning MX-style wheel emits many sub-notch events, each `ceil`ing to 1, so one flick fires the action dozens of times (rough with `exec`). Accumulate v120 to a whole notch before firing.

2. **Generated `docs/config.md` renders broken** — the new multi-line `config.reference.toml` comment produces a stray backtick + malformed bullet. Reword the comment so it regenerates cleanly.

Minor: notch bindings don't fire while fullscreen (that block only checks scroll triggers), and `source != Finger` also admits `Continuous`.

**Unstacking:** this commit only touches input/config with no dependency on the blur stack (#182/#185). Could you rebase it onto `main` so it lands independently?

```
git rebase --onto main perf/blur-scaling feat/wheel-notch-bindings
```

rollecode:
Done. Notch actions now work like keybindings, without exiting fullscreen, so the volume controls work over a fullscreen video. Another option would be to have it exit first, like the scroll bindings, but I preferred not to implement it that way. What do you think?

malbiruk:
I think it's the right thing, no need to exit fullscreen. Keybindings work in such way that if you do `spawn` action it doesn't exit fullscreen while `exec` does, so there already is flexibility. Canvas-related actions also exit fullscreen, so now it's properly handled. Merging.

PR #196, opened 2026-07-09
Add left_handed mouse config option

PulseRH:
## What

Adds a `left_handed` option under `[input.mouse]` that swaps left/right mouse buttons.

```toml
[input.mouse]
left_handed = true
```

## Why

libinput already exposes `config_left_handed_set()` for this — the same category of per-device libinput config as `natural_scroll`, which the mouse config already wires up via `config_scroll_set_natural_scroll_enabled()`. There was no equivalent for left-handed use, so this follows the exact same pattern end-to-end:

- `config/toml.rs` — raw `left_handed: Option<bool>` field
- `config/mod.rs` — resolves to `unwrap_or(false)`
- `config/types.rs` — `MouseDeviceSettings.left_handed: bool`
- `input/gestures/device_config.rs` — applies via `device.config_left_handed_set()`, gated on `config_left_handed_is_available()` since not all pointing devices support it

## Testing

Built and ran on real hardware (Fedora, udev backend), confirmed left/right click are swapped with `left_handed = true` set.

`cargo fmt --check` and `cargo clippy --release` both clean.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -540,6 +540,7 @@ impl Config {
                 ),
                 accel_profile,
                 natural_scroll: m.natural_scroll.unwrap_or(false),
+                left_handed: m.left_handed.unwrap_or(false),
             }
         };
 
--- a/src/config/toml.rs
+++ b/src/config/toml.rs
@@ -87,6 +87,7 @@ pub(super) struct MouseDeviceFileConfig {
     pub accel_speed: Option<f64>,
     pub accel_profile: Option<String>,
     pub natural_scroll: Option<bool>,
+    pub left_handed: Option<bool>,
 }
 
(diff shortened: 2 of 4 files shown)


malbiruk:
Thanks for the PR, one thing:

 `build & test` is red -- `every_config_field_is_documented`:
 
```
config fields defined in code but undocumented in config.reference.toml: ["input.mouse.left_handed"]
```
 
Add it under `[input.mouse]` in `config.reference.toml`, then regenerate the docs:

```toml
 # left_handed = false         # swap left/right mouse buttons
 ```
 
 ```bash
 UPDATE_CONFIG_DOCS=1 cargo test docs_config_md_is_up_to_date
 ```

`cargo test` catches both locally. Code itself looks good.

Upd: I'll do it in follow-up commit

PR #215, opened 2026-07-16
fix incorrect use of window_bbox instead of window_bbox_with_popups

SputerSnoot:
I noticed this issue when hovering over popups that are over another window - gtk closes them because the compositor focuses the window behind.

Tested fixes:
- Use window_bbox_with_popups instead of window.bbox() for detecting what's under the cursor; now popups work correctly and retain focus.
- Changed the method naming and documentation across the codebase because it was misleading and incorrect regarding its handling of popups.

Untested fixes:
- Fixed rendering visibility checks (`visible_rect.overlaps`) and frame callback throttling to use `window.bbox_with_popups()`. This resolves a hidden multi-monitor edge case where popups overhanging across screen boundaries would be visually clipped and suffer from severe frame throttling.
- Fixed `mark_dirty_for_surface` to correctly include overhanging popups when calculating which outputs need redrawing, preventing popups from visually freezing on adjacent monitors.

**(It just makes sense but i don't have a second monitor to test it, it would be really good to test for issues before and after the change)**

Holy vibe and slop! Is it Claude to blame? There's literally a comment describing that the method checks for windows and popups, and then it uses a method that doesn't include them. Its code should always be questioned.

I added the instruction to the caveats.md, i hope Claude reads it.

### DIFF
--- a/dev/docs/caveats.md
+++ b/dev/docs/caveats.md
@@ -4,7 +4,7 @@ Things to keep in mind as the codebase grows.
 
 ## Never touch `Space` directly — go through the stage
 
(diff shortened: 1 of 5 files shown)


malbiruk:
Thank you for the fix! It's me to blame of course, though that was a really big refactor I assigned to Claude. Additionally to caveats, i'll secure this fix with linting and regression test in follow up.

Issue #219, opened 2026-07-17
animate_blur overheats the GPU as frosted windows accumulate: non-bottom-most windows re-render the whole scene per refresh


rollecode:
With animate_blur on and two frosted terminals over an animated shader, the GPU sits at 43% / 136 W and 85+ C with fans pinned above 90%, at 3440x1440@144 on NVIDIA. Pausing the background drops it instantly, so it reads as an expensive wallpaper, but the background alone is 18% / 61 W: the blur multiplies it.

## Root cause

The shared-slice condition from #182 is a z-order test, not an overlap test:

```rust
behind_starts[i] >= background_start
```

Only the bottom-most frosted window passes. Every other one falls through to the per-window path and re-renders the entire scene into a full-output FBO on every refresh. On the canvas windows mostly sit side by side, so "above something in z-order, overlapping nothing" is the normal case, and the cost scales with frosted window count like pre-#182.

Instrumented counters, three non-overlapping frosted windows, animate_blur_fps = 20:

```
shared=19/s  slice=19/s  fallthrough=38/s
```

One slices, two re-render the scene 19 times a second each for identical pixels.

## Fix candidate

Fall through only when an element below the window, padded by the blur reach, actually intersects it. Overlapping windows keep the current behaviour. Testing it locally: the same layout runs 21% / 84 W with frost live.

rollecode:
The z-order check was only part of it. Two more paths bypass the throttle while panning: camera moves force a shared refresh every frame (full scene render + full-output blur, up to 144/s), and the window's screen rect is part of the background hash, so stacked windows recompute per pan frame on top of that.

Tried these:

- fall through to the per-window path on actual overlap (padded by the blur reach), not z-order
- occluded windows drop out of the animated cadence: their refresh is the scene re-render, which is the part that scales with window count
- camera-forced refreshes obey animate_blur_fps, and occluded recomputes hold during a pan, catching up 150 ms after settle. Canvas windows move with their backdrop, so the held frost stays correct while panning

Idle floor is back to 63-66 W with frost live. Spikes only during interaction, and opening more windows doesn't move the cost anymore: fans went down during a window-storm test.

malbiruk:
Thanks for the issue and the fix!

PR #220, opened 2026-07-17
Fix animate_blur cost scaling with window count: slice on actual overlap, throttle pan refreshes

rollecode:
animate_blur cost grows with frosted window count (#219): the shared-slice condition from #182 is a z-order test, so on a canvas where windows sit side by side every frosted window except the bottom-most re-renders the whole scene per refresh. Panning makes it worse twice over: camera moves force a shared refresh every frame instead of obeying the throttle, and the window's screen rect is part of the background hash, so stacked windows recompute per pan frame on top.

- fall through to the per-window path only when an element below the window (padded by the blur reach) actually intersects it; side-by-side windows all take the shared slice, overlapping ones keep the current behaviour
- occluded windows drop out of the animated cadence: their refresh is the scene re-render, which is the part that scales. Camera and geometry changes still update them
- camera-forced refreshes obey animate_blur_fps, and occluded recomputes hold while a pan is in flight, catching up 150 ms after settle. Canvas windows move with their backdrop, so the held frost stays correct while panning

Measured on the setup from #219 (3070, 3440x1440@144, animated shader, two frosted terminals, animate_blur_fps 20): idle floor 63-66 W with frost live vs 136 W before, interaction spikes transient, and opening more windows during a storm test moved nothing: fans went down while window count went up.

### DIFF
--- a/src/render/blur.rs
+++ b/src/render/blur.rs
@@ -514,6 +514,26 @@ pub(crate) fn process_blur_requests(
 
     // ── First pass: create/resize caches, update dirty flags, decide who recomputes ──
     let pad = blur_pad(blur_strength, blur_passes);
+
+    // behind_starts alone is a z-order test: side-by-side windows all read as
+    // "stacked" and lose the shared slice. Fall through only when an element
+    // below this window (padded by the blur reach) actually intersects it.
+    let elem_scale = smithay::utils::Scale::from(output_scale);
+    let occluded_by_lower: Vec<bool> = blur_requests
+        .iter()
+        .enumerate()
+        .map(|(i, req)| {
+            let mut probe = req.screen_rect;
+            probe.loc.x -= pad;
+            probe.loc.y -= pad;
+            probe.size.w += 2 * pad;
+            probe.size.h += 2 * pad;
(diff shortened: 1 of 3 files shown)


malbiruk:
Thank you! The occlusion reframing is the right fix, and the numbers speak for themselves. Two things before merge:

1. Blank frost when a resize lands during the hold: cache.resize() runs before the hold check and reallocates the textures uninitialized; the hold then skips the recompute but the blank cache.texture still gets inserted. Trigger: pinch zoom with overlapping frosted windows (per-frame resize + hold). Fix: capture resized before the resize call and add && !resized to the hold guard.
2. Pinned/Top/Overlay are screen-fixed, so the canvas slides under them while held — and a continuous pan holds them the whole time, not 150 ms. Suggest && !camera_dirty in the hold guard (it's computed right above), unless that case showed real heat in your testing.

rollecode:
Fixed in deafdaa.

On the heat question: nothing screen-fixed is frosted in my config (the blur rules only cover terminals), so that path never appeared in the measurements. No reason to hold them.

I have no touch devices to test with, so I personally haven't been able to test the pinch-resize path.

Issue #227, opened 2026-07-21
Animations


rywby-dot:
What about animations  #34 

https://github.com/user-attachments/assets/165a5a87-5b03-4c01-8940-76b0f42bf38f

Animations for:
- Open/close window
- zoom-to-fit, zoom-to-fit-snapped, fill-window
- pan-viewport, nudge-window
- zoom (with mouse wheel)
- fullscreen-toggle

Everything setup globally with
```
[navigation]
animation_speed = 0.2
```

You can check it here https://github.com/rywby-dot/driftwm/tree/animations

[diff.txt](https://github.com/user-attachments/files/30230895/diff.txt)

made with codex

malbiruk:
Thanks for this! The demo looks great, and I checked out the branch. I propose splitting it:

1. Camera half (`ZoomAnimationAnchor`, cursor-anchored wheel zoom, animated `pan-viewport`) I'd merge as is; please open a PR with just these changes (no config/docs or udev scheduler changes needed).
2. Window effects half (open/close/geometry/fullscreen): I'd adapt this myself with you as co-author. The close animation needs a redesign: xdg close is a request the client may decline, so unmapping before `send_close()` can permanently hide a window that shows a "save changes?" dialog, and client-initiated closes (the common case in a CSD-first compositor) bypass this path and get no animation. Snapshotting at actual unmap is the fix, and that half also collides hard with the big session/suspend branch (#105) landing soon, so I'll build it on top of that. The render-only transform architecture you picked is exactly right though, and I'll keep it.

Also I'd close this issue in favor of #34 as the tracking issue, please link the PR there.

Issue #34, opened 2026-04-14
Add animations for window states (open/close/move)


komarukomaru:
I suggest adding an animation for opening and closing windows (Fade in/out for ex). Maybe even a movement animation (like in Gnome).

malbiruk:
Shipped in **v0.16.0**, covering more than open/close: windows animate on open, close, move, resize, fit, fill and fullscreen, and the camera animates on pan and zoom. Tune with `[effects] animation_speed` (default `0.5`, `1` = instant) and `[effects] animation_scale` (default `0.95`, `1` = fade only).

Thanks to @rywby-dot, whose prototype in #227 is what both halves are adapted from.

PR #229, opened 2026-07-21
Camera animations

rywby-dot:
Animation to cursor-anchored wheel zoom and pan-viewport #34 #227 https://github.com/malbiruk/driftwm/issues/227#issuecomment-5036047560


malbiruk:
Thank you!

PR #240, opened 2026-07-27
Replace .unwrap() calls on seat accessors with graceful error handling

AvengerAnubis:
Two defensive fixes for config hot-reload and session-lock paths:

1. `state/reload.rs` - `reload_config_from_contents` called
   `seat.get_keyboard().unwrap()` twice when the keyboard layout or
   repeat rate changed. If the keyboard seat is absent during reload
   (possible during early init), this panics the compositor.
   Changed both to `if let Some(keyboard)` with a warning log and
   graceful fallback (keep the existing config).

2. `handlers/mod.rs` - `SessionLockHandler::lock()` called
   `seat.get_pointer().unwrap()`. Same panic risk if the pointer
   seat isn't available. Changed to `if let Some(pointer)`.

Both paths are guarded by equivalent checks elsewhere in the codebase;
this brings the reload and lock paths in line with the rest.

### DIFF
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -1186,24 +1186,22 @@ impl SessionLockHandler for DriftWm {
             self.loop_handle.remove(pending.timer_token);
         }
         let serial = smithay::utils::SERIAL_COUNTER.next_serial();
-        let pointer = self.seat.get_pointer().unwrap();
-        pointer.unset_grab(self, serial, 0);
+        if let Some(pointer) = self.seat.get_pointer() {
+            pointer.unset_grab(self, serial, 0);
 
-        // Deactivate any pointer constraint held on the current focus surface.
-        // Without this, a Wine game (or any client with a Locked constraint)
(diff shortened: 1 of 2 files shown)


malbiruk:
Thanks for looking at this! I don't think these two `.unwrap()`s can actually fire, though, as the seat gets its keyboard and pointer at startup in `DriftWm::new`, and nothing ever removes them, so by the time config reload or session lock runs they're always present. Reload also can't happen before startup finishes, so there's no early-init window either.

Also `.unwrap()` on these two accessors is the house style -- there are around forty of them across the codebase, so this makes three sites the exception rather than the rule.

I'm going to close this, but if you'd like to move the whole codebase off `.unwrap()` on the seat accessors -- a small `self.keyboard()` / `self.pointer()` helper applied everywhere -- I'd be glad to talk about that as its own change.

If you actually hit a panic here rather than spotting it by reading, please do reopen with the backtrace -- that would mean something else is wrong and I'd want to chase it.

PR #241, opened 2026-07-28
driftwm-session: fall back to direct exec when systemd is absent

AvengerAnubis:
The session wrapper script currently requires systemd user manager via systemctl. On runit-based distros (Void Linux) where systemd is not available, the script exits with an error.

Changes:
- Detect systemctl availability at runtime; if absent, run driftwm directly (the login-shell re-exec already sets up the environment)
- Makefile: only install systemd service files when systemctl is found on the build host

This makes the same driftwm.desktop / driftwm-session binary work on both systemd and non-systemd systems without configuration changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,13 +11,17 @@ TARGET_DIR = $(or $(CARGO_TARGET_DIR),target)
 build:
 	cargo build --release
 
+HAS_SYSTEMD := $(shell command -v systemctl >/dev/null 2>&1 && echo yes || echo no)
+
 install:
 	install -Dm755 $(TARGET_DIR)/release/driftwm $(DESTDIR)$(BINDIR)/driftwm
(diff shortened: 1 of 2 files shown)


malbiruk:
Thanks for this! The script side is right and I'd like to take it. It matches what the compositor already does in `src/main.rs`, where the systemd and D-Bus env exports are semicolon-separated so the D-Bus half still runs without systemd.

One change before merge: please drop the Makefile hunk. `$(shell command -v systemctl)` probes the build host, not the target, so packages built in minimal chroots would silently install with no unit files and break the session with no error. Two inert files under `$(LIBDIR)/systemd/user/` cost a runit system nothing, so unconditional install is fine (or add an explicit `INSTALL_SYSTEMD ?= 1`).

AvengerAnubis:
Done. Thank you too for making driftwm, I really enjoy using it!

PR #245, opened 2026-07-29
fix: render normal windows during Pending lock state, switch to lock frame only when fully Locked

AvengerAnubis:
# Prevents the screen flickering after locking.
## Before
After locking, the screen goes black for a very small (but noticable) period of time, because of the lock screen being prepared and not rendered yet (`is_locked()` is actually `!Unlocked` rather than just `Locked`). Nothing is being rendered -> screen is black.
## After
Renders the lock screen only when actually `Locked`
## Alternative
Change `is_locked()` behavior to return `true` when the session lock state is `Locked`, instead of it returning `true` at any state but `Unlocked`. But it seems that `is_locked()` is being used in many placed of the codebase, especially in input handling, so I decided to play it safe and this PR only hard-codes a specific behavior for lock screen rendering. Let me know if you prefer that `is_locked()` be changed instead

### DIFF
--- a/src/handlers/compositor.rs
+++ b/src/handlers/compositor.rs
@@ -240,10 +240,11 @@ impl CompositorHandler for DriftWm {
 
             let all_ready = self.lock_surfaces.keys().all(|o| ready_outputs.contains(o));
             if all_ready {
+                // Cancel the deadline timer BEFORE we move out of Pending.
+                self.cancel_pending_deadline();
                 let old =
                     std::mem::replace(&mut self.session_lock, crate::state::SessionLock::Unlocked);
                 if let crate::state::SessionLock::Pending { locker, .. } = old {
-                    self.cancel_pending_deadline();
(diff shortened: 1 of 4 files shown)


malbiruk:
Thanks for this, your diagnosis is right.

The blocker was upstream of your change: we sent `locked` before the lock frame was composited, which is harmless only because Pending blanks the screen. Remove the blank and a client that suspends on locked races the first lock frame, so unlocked content can show on resume. I've pushed the other half to main -- locked now waits for a lock frame to be page-flipped on every output.

Could you rebase on that and add a deadline (niri uses 1s) after which we blank anyway, since Pending is otherwise unbounded? It'll be a small rewrite rather than a clean rebase as the gate moved into `renders_lock_frame()`. Keep input blocked during Pending as you had it. Happy to merge once rebased.

AvengerAnubis:
here, I think it's done

PR #166, opened 2026-06-24
feat: add graphics tablet support and fix Nix development shell

dominicegginton:
Implement graphics tablet (digitizer) support via the `wp_tablet_manager_v2` Wayland protocol, enabling high-definition pen input and stylus hover tracking on the infinite canvas.

- Initialize `TabletManagerState` on the compositor state and delegate the protocol.
- Track physical graphics tablet devices and register tools on hotplug.
- Project absolute tablet hardware coordinates to screen coordinates and convert them to absolute world space using the infinite canvas coordinate transforms.
- Forward high-definition stylus inputs (tilt, pressure, distance, rotation) to native tablet-aware clients like Krita or GIMP.
- Emulate standard pointer buttons and motion for non-tablet-aware legacy apps and server-side window title/border decorations.
- Correctly route stylus hover proximity enter/exit states.
- Clean up Nix `devShell` definition in `flake.nix` by inheriting inputs from the default package using `inputsFrom = [...]` to guarantee all dev-tools build correctly.
- Update user-facing and developer documentation with tablet configuration details.
- [ ] user configuration for mapping a graphics tablet input to output

_see also: https://github.com/malbiruk/driftwm/issues/118_

PR #247, opened 2026-07-29
Tablet fix feat: add graphics tablet support (rebased #166)

### LINKED ISSUE #118: Will support for graphics tablets be added?
Will support for graphics tablets be added?
graphics tablets 
(e.g., XP-Pen, Huion, Wacom) 
regular tablet 
(Sunshine + Monlights, iPad, Samsung)
I was just looking for a window manager that would let me work comfortably with two monitors, and I really liked Drift WM, but I can’t draw on it using a graphics tablet.
I tried OpenTablet; in absolute mode, the tablet seems to be detected, but there’s no pen pressure.
I tried adding graphics tablet support myself using AI, but nothing worked. 

Translated with DeepL.com (free version)
Sorry for the silly question

Shlubka:
Rebased version of #166 (feat: add graphics tablet support) with fixes for CI, docs, and merge conflicts.

Original authors: @dominicegginton
Follow-up work, testing, and conflict resolution by @Shlubka

Closes #118
Related: #166

---
**Testing done:**
Tested on ThinkPad X1 Yoga with Wacom pen (USB 056a:5146). The following features work correctly:
- Pressure, tilt, eraser, side buttons, distance, hover.

**Known issues (will be addressed in subsequent PRs):**
1. Cursor offset (few pixels off from pen position).
2. Click-through on window borders (click passes to content instead of decoration).
3. Screen rotation not mapped (wlr-randr transform doesn't affect tablet input).
4. Flatpak app drawing artifacts (Rnote produces arrow shapes instead of smooth strokes).
5. Stylus button drag behaviour (canvas jumps when lifting and pressing quickly).

All CI checks should pass after this PR.


malbiruk:
Thanks @Shlubka, the tablet core (thanks @dominicegginton) is good and I want it in.

The rebase took the June `README.md` and `flake.nix` wholesale, reverting ~5 weeks of docs and dropping `nixosModules.default`, `checks`, aarch64 and `doCheck = false` -- drop both files, keep the one tablet bullet. Also drop `d8989c8` (map tablet to output): `[input.touch.tablet]` silently parses and does nothing, and it uses the mapped output's size with the active output's camera, so a second-monitor pen lands wrong. Better as its own PR reusing the resolver `[input.touch]` already has.

Three of your five: 2 is tip-down sending a raw pointer button instead of going through `on_pointer_button`, so it never reaches the decoration handler or focuses the window. 3 is shared with touch -- neither reads `output.current_transform()`; two lines in its own commit if you're up for it, since you're the only one who can test a rotated panel with pen and touch. And one more: the tablet path never clears `hidden_by_touch`, so after any screen touch the pen drives an invisible cursor.

The dot cursor needs to yield to `grab_cursor`/`decoration_cursor` -- right now hovering a resize border with the pen shows a dot instead of the resize arrow -- and `dot_frames()` ignores output scale,

So: drop README/flake and `d8989c8`, `cargo fmt`, route tip-down through `on_pointer_button`, clear `hidden_by_touch`, guard the dot. Then it's mergeable.

Shlubka:
Updated the PR with the requested review changes. Apologies for the week-long delay.

Here’s what’s been addressed:
- `fix(tablet): route tip-down...` — Pen clicks are now routed through `on_pointer_button`.
- `fix(tablet): clear hidden_by_touch...` — The cursor is now correctly restored after touch input.
- `fix(input): respect output transform...` — Added support for screen rotation transforms.

I wasn't able to run more extensive testing on devices with extra features since the device owner wasn't available, but I did verify the core functionality on my Thinkpad Yoga.

malbiruk:
Thanks for rebasing this!

CI is red on `docs_config_md_is_up_to_date`: `docs/config.md` is generated from `config.reference.toml`, and `[input.tablet] map_to_output` documents a key that doesn't exist yet -- either add the option or drop the section.

Bigger one: the pen resolves its output through `active_output()`, so on multi-monitor it follows wherever the mouse last was. Touch does this per-device via `touch_output_for_device` plus `[input.touch] map_to_output`, worth the same shape here. Single-output doesn't hit it, which is why it tested clean.

Rest is small: #265 landed so the `input/touch.rs` transform change can go, `render/cursor.rs` can be reverted (that `if` already returns, so the `else` is just a reindent), and some commented-out code and a non-English comment are left in.

Also a guess -- the tip synthesizes a `wl_pointer.button` on top of the tablet event, so tablet-aware apps get two input streams. Probably your known issue 4?

Shlubka:
Thanks for the detailed review! I've implemented all the requested changes and pushed them. Here is a summary of what was done and tested.

### What was done
* **Per-device output resolution**: Replaced `active_output()` with `tablet_output_for_device`. It resolves the output via `[input.tablet] map_to_output`, libinput tags, physical size matching, and internal panel fallback, exactly mirroring the touch logic. Added `TabletSettings` to `InputConfig` and updated `config.reference.toml` / `docs/config.md`.
* **Local camera/zoom**: The canvas coordinates are now computed using the resolved output's own camera and zoom (`output_state(&output)`), rather than the global `self.camera()`. This prevents the multi-monitor "pen lands wrong" bug and notably fixed the Rnote drawing artifacts (Known issue #4).
* **Code cleanup**: Dropped the redundant transform code in `touch.rs` (thanks to the merged #265), reverted the `else` reindent in `render/cursor.rs`, and removed the commented-out code / non-English comments in `tablet.rs`. Also deduplicated the libinput helper functions by making them `pub(crate)` in `touch.rs` and importing them in `tablet.rs`.

### Testing done
Tested on ThinkPad X1 Yoga with Wacom pen (USB 056a:5146):
* Touchscreen and pen input work correctly. Screen rotation works for both.
* Pen `tip-down` correctly clicks decorations, resizes windows, and focuses.
* Pressure work in native and Flatpak apps. **Rnote artifacts are gone!**
* *Note on multi-monitor*: I couldn't test the exact multi-monitor scenario locally right now, but the logic follows the touch path exactly, using the resolved output's size, camera, and zoom.

### What remains (deferred to subsequent PRs)
* **Stylus button drag behavior**: Canvas jumps when lifting and pressing quickly.

PR #252, opened 2026-08-01
fix: Re-seat pointer focus when a layer under the cursor is destroyed

AvengerAnubis:
## Before
`pointer_over_layer` was always (unconditionally) being reset to `false` if any layer was destroyed. It was being updated by `focus_cascade` on mouse movement, so the state was frozen as false until next mouse movement.

### Case with Noctalias bar
https://github.com/user-attachments/assets/eed496cc-6277-4705-878a-cce6ed24c743

## After
Instead of resetting `pointer_over_layer` we are refreshing it via `refresh_pointer_focus()`

### Case with Noctalias bar
https://github.com/user-attachments/assets/d2830ccc-e29d-4453-aee9-eff7ba5b1ef1

### Filter inside of `layer_surface_under`
Skip layers with the `LayerDestroyedMarker` in hit test

### Fullscreen behavior fix
`enter_fullscreen` sets `pointer_over_layer` to `false`, and `exit_fullscreen_on` was restoring everything (camera, zoom, etc.) but `pointer_over_layer` - it was left as `false`. Now it refreshes it via `refresh_pointer_focus()`

### Regression test
src/tests/layer_destroy_focus.rs - test focus after destroying the layer the cursor was pointing at

### DIFF
--- a/src/handlers/layer_shell.rs
+++ b/src/handlers/layer_shell.rs
@@ -202,14 +202,10 @@ impl WlrLayerShellHandler for DriftWm {
         // smithay's pointer focus are only refreshed by real pointer motion, so
         // without this a layer destroyed under a stationary cursor would route
         // the next press/scroll to the canvas instead of the layer surface still
-        // beneath it. `layer_surface_under` skips the role-destroyed surface via
-        // the marker above, so the recompute lands on whatever is genuinely under
-        // the cursor. No-op while locked; `unlock` re-seats pointer focus anyway.
-        if self.session_lock.is_locked() {
-            self.pointer_over_layer = false;
-        } else {
-            self.refresh_pointer_focus();
-        }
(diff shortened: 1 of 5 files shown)


malbiruk:
Thanks, real bug and the right fix. I merged it onto main locally and confirmed the test fails without the `layer_destroyed` change.

Needs a rebase though: main now unmaps the layer immediately in `layer_destroyed`, which makes the `layer_surface_under` filter unreachable (I removed it and the test still passed), so please drop that hunk and its two imports. Could you also add a test for the `exit_fullscreen_on` half? 

Minor: `refresh_pointer_focus` already returns early when locked, so the else branch is redundant.

AvengerAnubis:
done

malbiruk:
Thanks!

PR #253, opened 2026-08-03
feat: Add 'toggle-touchpad' action support

AvengerAnubis:
### Usage:
- driftwm msg action toggle-touchpad -- toggles the touchpad state between disabled and enabled
- driftwm msg action toggle-touchpad on -- enables it
- driftwm msg action toggle-touchpad off -- disables it

### DIFF
--- a/config.reference.toml
+++ b/config.reference.toml
@@ -341,7 +341,7 @@
 # #   toggle-pin-to-screen    — pin/unpin the focused window to the screen (ignores pan/zoom, floats above)
 # #   reload-config           — hot-reload config file
 # #   toggle-cursor-pan       — toggle cursor edge-pan (see [navigation.edge_pan])
-# #   toggle-touchpad [on|off] — set the built-in touchpad enabled/disabled, or toggle (libinput send-events mode)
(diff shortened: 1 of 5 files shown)


malbiruk:
Thanks, this looks good and I'd like to land it.

One thing before it does: the name. `toggle-touchpad on` reads a little odd, the enum behind it is SetTouchpad, and the config calls these devices "trackpad" everywhere else (`[input.trackpad]`). Action names are public API and renaming later costs a migration entry, so I'd rather settle it now. How do you feel about `set-trackpad on|off|toggle`?

I'll add the config side as a follow-up: `[input.trackpad]` `send_events`, so the setting has a startup default and survives a VT switch.

AvengerAnubis:
I agree honestly, I left it as it was in the first commit when only actual toggle was implemented, but now as it can do on|off|toggle, yes, set-trackpad feels more suitable. Also I didn't notice that it's called "trackpad" elsewhere, I'm used to the word "touchpad" more. I'll change it to set-trackpad real quick. Thanks

AvengerAnubis:
done (rebased too)

PR #255, opened 2026-08-04
fix: session store live windows, preventing the race between the compositor and it's clients

AvengerAnubis:
## Problem
On logout the compositor and its clients receive SIGTERM at the same time. The session file can end up empty (`entries: []`) so the next login restores an empty desktop even though `restore_windows = true` is set and the previous session had windows open.

## Root cause
A race at session teardown. SIGTERM arrives via signalfd, whose handler calls `LoopSignal::stop()`. But calloop only checks the stop flag after dispatching the entire batch of events ready in the current poll iteration. The client-disconnect events (wayland socket data) sit in that same batch, and when they dispatch first they walk `toplevel_destroyed` -> `unmap_window` -> `stage.remove`, before the loop stops. `serialize_session_on_shutdown()` then runs against an already-drained stage and overwrites the previous good save with an empty entries list.

## Fix
Make the durable file reflect the live session continuously, so no single shutdown-time write is the source of truth:
1. Steady-state flush now records live windows. `session_store_flush()` writes with `include_live = true`, so the 1s debounced write persists live windows (as `Quit` records, subject to the same `restore_windows` resolution and widget/dialog filters as the shutdown write) alongside suspended stand-ins, carried-forward entries, cameras and bookmarks. A crash or SIGKILL now leaves the current canvas in the file.
2. Arm the debounce from every durable live mutation: window create (`new_toplevel`), window close (`toplevel_destroyed`), end of a move grab, end of a resize grab, the commit-time resize reposition (`handle_resize_commit`), and msg move.
3. The shutdown serialize stays as the final fsync'd stroke - unchanged behavior for graceful exits.

## Note
The count-matched carried-quit dedup now also applies at flush time (not only at shutdown). With `restore_windows` on, a carried record is dropped once its app relaunches live, and a window the user closes manually leaves the file - previously a carried record survived a manual close until the next graceful shutdown and resurrected the window. This matches the documented contract (manually closed windows are not restored).

## Rebase
After rebasing on updated main, re-checked the new code against the same contract:
- The new `msg resize` verb and grow/shrink actions never arm `ResizeState` - they write a `PendingResize` + plain sized configure - so their answering commit fell through the grab-path settle in `handle_resize_commit`. Now the Idle branch marks the session dirty when a live `PendingResize` answers.
- Nudge, go-to-bookmark, move-to-bookmark and send-to-output marked the store dirty only for stand-ins; live-window arms now mark too.
- Session tests updated to the frame-space file convention (commits [f87a352](https://github.com/malbiruk/driftwm/commit/f87a3520402e6aabf8c50022f1b855e78453fd2f) moved entries to visual-frame rects).

### DIFF
--- a/src/grabs/move_grab.rs
+++ b/src/grabs/move_grab.rs
@@ -455,9 +455,12 @@ impl PointerGrab<DriftWm> for MoveGrab {
         // Both arms armed `interactive_move` at grab install (guarding relaunch
         // adoption and animation starts); balance it here.
         data.disarm_interactive_move(&self.target);
-        // A stand-in's settled position (including a cross-output teleport) is
-        // durable — persist it on the session-store debounce.
-        if matches!(self.target, ClusterMember::Suspended(_)) {
+        // A settled position (including a cross-output teleport) is durable —
+        // persist it on the session-store debounce, live or stand-in alike.
+        if matches!(
+            self.target,
+            ClusterMember::Client(_) | ClusterMember::Suspended(_)
+        ) {
             data.session_store_mark_dirty();
         }
         // A pick-mode promote is the only move that sets grab_cursor (title-bar
--- a/src/grabs/resize_grab.rs
+++ b/src/grabs/resize_grab.rs
@@ -357,6 +357,9 @@ impl PointerGrab<DriftWm> for ResizeGrab {
                 // hands back a view move this grab held off — the stand-in arm's
                 // disarm does that for itself.
                 data.flush_deferred_views();
+                // The settled size (and the commit-time reposition that follows
+                // it) is durable — persist it on the session-store debounce.
+                data.session_store_mark_dirty();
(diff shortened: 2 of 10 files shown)


malbiruk:
Merging this. The steady-state flush is the right change on its own: the file now tracks the canvas continuously instead of resting on a single write at exit, so a crash or SIGKILL recovers your windows.

Retitling to "keep live windows durable at steady state" though, because I don't think it closes the logout race yet. `serialize_session_on_shutdown` still rebuilds from `stage.windows()`, so a drained stage writes an empty file over the good one. I tested the scenario from your description and it still lands at 0 entries.

For the follow-up I'm leaning toward dropping the shutdown write entirely, since your change makes it redundant. I'm going to take the follow-up myself since it reaches into the camera and animation tick, and I want it in the next release.

PR #260, opened 2026-08-12
Fix VRAM exhaustion with the monitor off: skip placeholder outputs in the idle frame-callback fallback

rollecode:
Unplugging the last output keeps a virtual placeholder so `active_output()` stays `Some`, but the placeholder has no DRM surface. `send_frame_callbacks_fallback` filtered only `dpms_off_outputs`, so the idle heartbeat kept servicing that placeholder: every client redrew once a second into buffers nothing composited and nothing released. On DisplayPort this fires on an ordinary overnight screen-off, since powering the monitor off drops the link and reads as a real disconnect rather than DPMS-off.

- filter `disconnected_outputs` out of the fallback's output list, the same predicate `refresh_foreign_toplevels` and `refresh_ext_workspaces` already use in this file. With every output a placeholder the list is empty and the heartbeat goes quiet
- regression test asserts silence for both a layer surface and a window while only placeholders remain, and that reconnecting restores the heartbeat

Measured on a 3070, 3440x1440 DP-1, nvidia-open 610.57.04: the compositor grew 278 MiB to 6483 MiB in 32 minutes with the monitor off, filled all 8 GiB of VRAM, then spilled to system memory overnight until 48 GiB of RAM ran out and the session died. Three mornings in a row ended in "No Signal" and a forced power cycle.

After the fix, a 67-hour screen-off: 259 MiB to 427 MiB total, peak VRAM 985 MiB of 8192, zero sysmem spill events across ~1750 samples. The remaining growth is a separate, self-correcting path (a shell's layer surfaces commit on their own timer, and a layer commit marks its output dirty unconditionally) and it is fully reclaimed on reconnect: 427 MiB dropped back to 142 MiB in a single sample when the monitor came back.

The test was written first and confirmed failing on unmodified `main` before the fix. `cargo fmt --check`, `cargo clippy --all-targets`, and `cargo test -- --include-ignored --skip soak` all pass.

### DIFF
--- a/src/render/lifecycle.rs
+++ b/src/render/lifecycle.rs
@@ -364,7 +364,16 @@ pub fn post_render(state: &mut crate::state::DriftWm, output: &Output) {
 /// doesn't gate them.
 pub fn send_frame_callbacks_fallback(state: &mut crate::state::DriftWm) {
     let time = state.start_time.elapsed();
-    let outputs: Vec<Output> = state.space.outputs().cloned().collect();
+    // Skip the virtual placeholders a full disconnect leaves behind: they have
+    // no DRM surface, so nothing composites and nothing releases the buffers a
+    // serviced client draws into. With every output a placeholder this leaves
+    // no output at all and the heartbeat goes quiet, which is the point.
+    let outputs: Vec<Output> = state
+        .space
+        .outputs()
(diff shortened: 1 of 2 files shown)


positiveway:
Please review and merge @malbiruk. This is a crucial fix.

